PR #10, opened 2020-12-26
get rid of test file

sdtblck:


sdtblck:
did some other basic cleanup, removed enwik8 from the repo so it's downloaded on demand, moved util functions to utils, etc.

StellaAthena (commented):
Looking at the changes to `trainenwik8.py` I’m a little confused by the fact that so many imports were removed despite the code not changing much. Was that just a lot of functionality that wasn’t actually being used?

sdtblck:
@StellaAthena imports were leftover from the old code that wasn't using deepspeed

StellaAthena (approved):

PR #16, opened 2020-12-27
add sparse attention support, with ability to specify at which layers…

lucidrains:
… to use sparsity


sdtblck (approved):
lgtm. In the future we should make the attention config more flexible (like in gpt-neo) but this is perfect for now :)

PR #17, opened 2020-12-27
add tfrecords dataset and make minor changes to configs/train script

sdtblck:


lucidrains (dismissed):

scripts/kill.sh
```diff
@@ -0,0 +1 @@
+pkill -f "python -u train*"
```
lucidrains, line 1: nice!

lucidrains (approved):

PR #21, opened 2021-01-01
fix small bug where sequence length is not passed into attention class

lucidrains:


StellaAthena (commented):
I tried this code, and it wasn't enough to get the code running on the server. See https://github.com/EleutherAI/gpt-neox/issues/22

StellaAthena (approved):
This allows gpt3small to run, but does not fix the problems with sparse attention (see https://github.com/EleutherAI/gpt-neox/issues/22)

PR #25, opened 2021-01-04
make mask value smaller by factor of 2

lucidrains:
because the mask is now added rather than used as a boolean fill, a max negative value for the mask may lead to overflow. decrease by factor of 2


StellaAthena:
Discussion in #gpt-neo indicates that it’s ambiguous if this helps with the problem people were having. Waiting to approve until it’s clearer why some people’s code isn’t working.

StellaAthena (approved):
This is a useful change but does not fix the NaN issue.

PR #34, opened 2021-01-04
feedforward GLU on by default

lucidrains:


sdtblck (approved):

hendrycks:
"Good to know T5 is doing that
They are also going the FF Glu [variant] route. We should just have that on by default" - lucidrains
That might be jumping the gun. The second author of T5 1.1 was Noam, who wrote the GLU variant paper where he combined the GELU with the GLU. Not exactly an organic design choice adopted the broader community.
T5 1.1 https://arxiv.org/pdf/2002.08910.pdf
Noam's paper: https://arxiv.org/pdf/2002.05202.pdf

PR #35, opened 2021-01-04
Fix error in extracting OWT2 dataset

steven-mi:


lucidrains (approved):

lucidrains:
> Last pull request response was too fast

np lol, nothing mission critical :+1:

PR #39, opened 2021-01-05
Add improved data downloading class / pipeline

sdtblck:
- run prepare_dataset.py with your model args to download the required dataset (if you run with deepspeed it downloads once for every rank)
- to add a new dataset, create a class that inherits from DataDownloader and add it to the DATA_DOWNLOADERS registry


sdtblck:
Have now added the dataset downloading to the main train scripts so you don't have to run a separate python file.
iirc torch.distributed.barrier() has a timeout of 30 mins, so we need to figure out how to increase that in the case that the data takes longer to download

sdtblck:
Also removed adam optimizer if using deepspeed - passing in an optimizer override's deepspeed's ZeRO optimizer, which we want to use

PR #51, opened 2021-01-09
Update data_utils.py

ShivanshuPurohit:
Change the if statement in `get_tokenizer`, as the first if sets the `tokenizer_type` to `None` and errors at `.lower()` method


StellaAthena:
In your second commit yourself change the conditional statement to check for `"hf_gpt2tokenizerfast"`. However you haven’t restored the actual use of `GPT2TokenizerFast`. What’s the motivation for this change? Was it deliberate?

ShivanshuPurohit:
I tried the two changes and the second one seems to work for some reason. With the first, neither was the pbar progressing nor was anything being written into the tfrecords folder.

sdtblck (approved):

sdtblck (approved):

PR #65, opened 2021-01-15
Hub

raijinspecial:
This branch adds a dataset class that reads directly from a zarr array stored and accessed via activeloop hub. This is for showing how to load batches from a single large array stored remotely and get a gauge of the performance.  The data here was taken from a pretokenized 100k line tfrecord generated from 00.jsonl.zst.

bash scripts/train_hub.sh should start training, tested on single machine only.


StellaAthena (changes requested):
This appears to be broken, possibly due to changes that have been made to the upstream fork that haven't been integrated. Please integrate with the upstream fork and resubmit.

StellaAthena:

PR #77, opened 2021-01-23
Implement distributed training using Kubernetes

leogao2:
Distributed training should work without any problems now.

To use:

1. Push to docker (todo: automate this step using github pipelines)
2. Run `./deploy_k8s.sh` - this should create all of the containers necessary and drop you in a shell. Optional: change number of nodes in kubernetes/deploy_k8s.yml to however many you want
3. Run your deepspeed command (i.e `deepspeed --hostfile=hosts train_enwik8.py --deepspeed --deepspeed_config configs/deepspeed_zero2.json`) in the shell. Make sure to include `--hostfile=hosts`.


leogao2:
Todo: set up eleuther docker hub account and make pipeline for building containers

Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM atlanticcrypto/cuda-ssh-server:10.2-cudnn
```
amannm, line 1: Looking at this dockerfile, https://github.com/coreweave/cuda-ssh-server/blob/master/Dockerfile i can see that it isn't too complicated, you might want to consider just pasting that in here so the project isn't more associated with your current vendor than it needs to be.

Plus it looks like there might be some some default passwords exposed there.

leogao2: I guess we could do that at some point, maybe worth creating an issue for, but I don't think it's high priorty.

As for default passwords, actually, what we do is we disable authentication entirely (since it's all on a private network). Suboptimal, yes, but it works for now and we can figure it out later, concurrently with figuring out actual training.

StellaAthena: > Plus it looks like there might be some default passwords exposed there.

Not quite. When you create a deployment you can check the logs to see the password. After doing so, the logs are flushed.

Is it the most secure thing in the world? No
Is it a problem? I don't think so

kubernetes/deploy_k8s.yml
```diff
@@ -0,0 +1,57 @@
...
+      terminationGracePeriodSeconds: 10
+      containers:
+      - name: neox
+        command: ["/usr/sbin/sshd"]
+        args: ["-D"]
+        tty: true
+        image: leogao2/deepspeed_eleuther
+        ports:
+          - name: sshd
+            containerPort: 2222
```
amannm, line 26: shouldn't this be `22` as suggested by the `EXPOSE` at https://github.com/coreweave/cuda-ssh-server/blob/master/Dockerfile#L37

StellaAthena: That's what I had thought as well, but Leo and I tested it and it seems to work with 2222. Not sure why / what difference it makes though.

StellaAthena (approved):

PR #84, opened 2021-01-23
Update requirements.txt

srulikbd:
it seems that without a docker environment, this package is missing..

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ transformers
 tensorflow==2.4.0
 ftfy
 lm_dataformat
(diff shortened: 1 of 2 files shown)


StellaAthena:
Strange, as we only got the docker environment working last night. I’ll test this when I get home.

sdtblck:
ah, we only used this for initial testing, best to just take it out of the code

sdtblck (approved):

srulikbd:
thanks!
should I prefer running neox inside a docker?

PR #87, opened 2021-01-23
Batch size needs to be specified

joshlk:
Small change so `config/gpt3_small.json` works


StellaAthena (approved):
I guess if you don't use the ZeRO optimizer, yes this is needed. Do keep in mind that the ZeRO optimizer ignores this and uses the value in the deepspeed config file.

PR #89, opened 2021-01-24
Fix train pipeline

sdtblck:
Fix pipeline training scripts.

Some fixes are *literally in the deepspeed library* so setup will need to install our deepspeed fork https://github.com/EleutherAI/DeepSpeed to properly work.


StellaAthena (approved):
Works for several people locally and on the CW servers! A big win 👍

PR #91, opened 2021-01-24
Miscellaneous docker QoL improvements

leogao2:
From @StellaAthena:

> 1. Store the hostfile in /job/hostfile so that the code automatically finds it and we don't need to provide it as an argument.
> 2. Install DeeperSpeed instead of DeepSpeed
> 3. The script lands you in ~/app/, which contains the contents of the GPT-NeoX repository. Would it be possible to have the GPT-NeoX repo be one layer deeper, so that it drops you in ~/app/ which contains the directory gpt-neox? That would be very pleasant but isn't necessary.

Addresses 1 & 2, fixes XY behind 3

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ RUN apt-get update && \
     pip3 install torch pipx && \
     python3 -m pipx ensurepath
 
-RUN mkdir -p ~/.ssh /app && \
+RUN mkdir -p ~/.ssh /app /job && \
     echo 'Host *' > ~/.ssh/config && \
     echo '    StrictHostKeyChecking no' >> ~/.ssh/config && \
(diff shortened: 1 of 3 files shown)


StellaAthena (changes requested):
This creates a `job` directory and puts the host file in it, but the directory is not in `app`! I believe this is fixed by changing line 10 to `kubectl cp $PWD/hostfile $MASTER_ID:/app/job`.

StellaAthena:
Also, having to download and tokenize enron is annoying. I can't imagine how annoying OWT2 or the Pile will be. Can we set up a persistent storage space that doesn't get cleaned up and gets accessed by the new instances we create? We can load the data onto that and therefore only have to download and tokenize once.

leogao2:
I double checked with deepspeed documentation, `/job/hostfile` is the correct path.

StellaAthena (approved):

PR #102, opened 2021-01-27
Update deploy_k8s.sh

leogao2:


joshlk:
Hi @leogao2 , can you pleaes add the `leogao2/gpt-neox` `DockerFile` to the repository? I can't find it anywhere

joshlk:
Do you think its worth setting up a CI pipeline that builds the docker image uploads it to a `EleutherAI` account. Otherwise each time we change the code we will need to prompt you @leogao2? I have done this before and can help

leogao2:
@joshlk Agree that we should do that eventually, this is just a stopgap. If you want to work on a CI pipeline that would be great.

Also, another problem with CI pipelines is that the build could potentially take quite some time, which is not good for fast iteration.

StellaAthena (approved):

PR #88, opened 2021-01-23
Added MPU from Sid's MegatronPipeline

glebshevchukk:
Issue #75: Added mpu code that Sid wrote for https://github.com/EleutherAI/MegatronPipeline and update train.py file to use it for model building. Only tested with a single machine / GPU so I'm not sure if it works correctly across machines. 

Also includes changes for Issue #40 that include the previous dataset for TFRecords and another one for loading directly from Json. 

To run, try 
``` sh scripts/train_gpt3small.sh```

PR #103, opened 2021-01-29
Adding MPU training & generation code

glebshevchukk:
(Extending from #88) Tested MPU model more using a single GPU, did not push pipeline code for training/generation but running the below should work for a single computer :

```
sh scripts/train_gpt3small.sh
sh scripts/generate_gpt3small.sh
```


StellaAthena:

PR #105, opened 2021-01-29
Docker build action test

joshlk:
Github action to build and push docker image.

I have tested it on my [fork](https://github.com/joshlk/gpt-neox/runs/1791909747?check_suite_focus=true) so that I can add the github secrets required to push to DockerHub to the repo setting. The build and push works. Annoyingly it takes around 12 mins each time as it's not caching the layers and starting from 0 each time. There might be an easy way to enable this which should speed it up.

The build time might not matter as we can configure it so this action doesn't block merging PRs.

I also removed `git remote set-url origin https://github.com/EleutherAI/gpt-neox/` from the DockerFile as it doesn't work on my machine locally or within the github action process. If its needed we can add to the k8 def file


StellaAthena:
@leogao2 @sdtblck how do we want to handle credentials for this? Can we give our GitHub accounts push authority to Docker?

joshlk:
You can add the credentials to the repo or organisation and only "owners" can modify them: https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository

I suggest we push docker images to a EleutherAI account. So the docker image name would be EleutherAI/gpt-neox:<git-sha>. We can then references the images using the git-shar tag in the k8 deployment spec

StellaAthena:
> You can add the credentials to the repo or organisation and only "owners" can modify them: https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository
> 
> I suggest we push docker images to a EleutherAI account. So the docker image name would be EleutherAI/gpt-neox:git-sha. We can then references the images using the git-shar tag in the k8 deployment spec

This seems quite reasonable to me.

joshlk:
This is working now and ready to be merged

joshlk:
Merged into PR #107

PR #113, opened 2021-02-03
merged script for train.py, train_pipeline.py

ShivanshuPurohit:
Not tested yet.


StellaAthena (changes requested):
Can you re-implement this with pipeline being specified as an entry in the JSON file rather than as a command-line argument?

ShivanshuPurohit:
Sure. Can you tell me what the command should look like with JSON instead
of the --mode flag?

StellaAthena:

PR #118, opened 2021-02-13
New codebase based on megatron

StellaAthena:
This PR replaces the old GPT-NeoX codebase with the Megatron codebase.


leogao2:
Non squashed version is at #119

PR #119, opened 2021-02-13
Replace neox code with megatron

leogao2:


StellaAthena (commented):
I went through and changed the kube files to point at gpt-neox again. I **think** it should work but someone should test to confirm this before approving.

StellaAthena (dismissed):
Works right out of the box!

StellaAthena:
I added legal compliance stuff.

Also, before this PR is merged we need to double check that the HOSTFILE is being created correctly. While the code ran out-of-the-box for me, it did **not** run in parallel. I only just noticed this unfortunately. I am leaving my review as approval so that it's non-blocking, but **please double check this before merging**.

StellaAthena (approved):

PR #132, opened 2021-02-21
Clean up Neox configuration

joshlk:
Clean up neox configuration so config files can be used instead of a mishmash of files, command line args and enviroment variables.

Aim:
* All parameters can be set using passed json files
* No parameters are repeated
* Modify megatron's codebase as little as possible to make it easier to merge upstream megatron changes in the future.

Nice to haves:
* [JSON schema files](https://json-schema.org/)
* Single config documentation

Todo:
- [x] Convert all examples and configs to new configuration
- [ ] Create config documentation with all possible parameters
- [x] Separate configs into: model and system
- [x] Cast numbers to numbers in JSON (suggested by @StellaAthena)
- [x] Calculate batch size from other parms (`micro_batch_per_gpu*GAS*n_gpus`)


joshlk:
Created a draft implementation. Example usage:

```
./deepy.py pretrain_gpt2.py -d configs ds_pretrain_gpt2.yml eleutherai_cluster.yaml
```

* `configs` is the directory the config files are in
* `ds_pretrain_gpt2.yml`, `eleutherai_cluster.yaml` are two JSON-ish (ok they are YAML but its basically JSON with comments) files. `deepy` takes all config files and merges then (so we can separate out the parameters how we want). Currently `eleutherai_cluster.yaml` has system stuff and `ds_pretrain_gpt2.yml` is to do with the model

The parameters in the yaml files are automatically separated into the DS runner, Megatron and DS config file parts. They are then converted into the "old" format and provided to the scripts. I wanted to do it this was so that I made as little changes as possible to the megatron/deepspeed codebase - making it easier in the future to merge upstream changes.

Some parameters are also automatically derived, such as the megatron "fp16" param from the DS runner "fp16" param.

configs/ds_pretrain_gpt2.yml
```diff
@@ -0,0 +1,64 @@
+{
+   "num_nodes":"1",
+   "num_gpus":"8",
+   "model-parallel-size":"4",
+   "num-layers":"24",
+   "hidden-size":"1024",
+   "num-attention-heads":"16",
+   "seq-length":"1024",
+   "max-position-embeddings":"1024",
+   "batch-size":"9",
```
ShivanshuPurohit, line 10: Why 9?

joshlk: It’s taken from the corresponding examples script

joshlk:
```
./deepy.py pretrain_gpt2.py -d configs ds_pretrain_gpt2.yml eleutherai_cluster.yaml
```

The example above exactly replicates the parameters used in `examples/ds_pretrain_gpt2.sh` - try it for yourself. It is not intended to show all possible configurations. I can create such a config later. @ShivanshuPurohit I am going to undo your commit as `pipe-parallel-size` isn't used in the original example.

ShivanshuPurohit:
I see. No problem. I thought all the keys were supposed to be initialized.

configs/ds_pretrain_gpt2.yml
```diff
@@ -1,42 +1,42 @@
...
+   "data-impl": "mmap",
+   "split": "949,50,1",
+   "distributed-backend": "nccl",
+   "lr": 1.5e-4,
+   "lr-decay-style": "cosine",
+   "min-lr": 1.0e-5,
+   "weight-decay": 1e-2,
+   "clip-grad": 1.0,
+   "warmup": 0.01,
+   "checkpoint-activations": True,
```
joshlk, line 27: @StellaAthena `true` is how true is represented in JSON. YAML allows for either `True` or `true` but lets just use `true/false` for now

megatron/training.py
```diff
@@ -252,6 +254,19 @@ def setup_model_and_optimizer(model_provider_func):
...
+    # If JSON then directly load it
+    deepspeed_conf = None
+    if hasattr(args, 'deepspeed_config'):
+        deepspeed_json_conf = args.deepspeed_config
+        if len(deepspeed_json_conf) > 2 and deepspeed_json_conf[0] == "'" and deepspeed_json_conf[-1] == "'":
+            deepspeed_json_conf = deepspeed_json_conf[1:-1]  # Remove shell quotes
+        try:
+            deepspeed_conf = json.loads(deepspeed_json_conf)
+            args.deepspeed_config = None
+        except JSONDecodeError:
```
sdtblck, line 267: Should we just except this here? What if we are trying to read a json, but there's a parsing error?

sdtblck (commented):
Looks great to me!

Can't see any problems aside from the one comment i made

megatron/training.py
joshlk: I wanted to add this here so if the user used the "old way" of defining a DS conf by providing a file path to a JSON then it would still work

joshlk:
Ok im going to move forward and convert the other configs and features...

joshlk:
Couldn't convert `examples/ds_pretrain_gpt2_pipe_flops_profile.sh` as `ds_config_flops_profile.json` doesn't exist.

configs/ds_pretrain_gpt2.yml
StellaAthena: Hmm interesting. I had assumed consistency with Python was preceded, but if JSONs use a different standard I’m happy to use what’s consistent with that.

joshlk:
@sdtblck @StellaAthena @ShivanshuPurohit its now working. For example the following should work with any cluster size:

```
./deepy.py pretrain_gpt2.py -d configs ds_pretrain_gpt2_medium_pipe.yml eleutherai_cluster.yml
```

Can we work out what configs we actually want and are using and scrap the others. I also am going to generate some docs of the parameters so we can add it to the a readme. Otherwise its ready to merge.

StellaAthena:
This was working and apparently something I did broke it. Currently getting the error
```
deepy.py: error: unrecognized arguments: --model-parallel-size --pipe-parallel-size 2 --num-layers 24 --hidden-size 1024 --num-attention-heads 16 --seq-length 1024 --max-position-embeddings 1024 --batch-size 4 --gas 16 --train-iters 320000 --lr-decay-iters 320000 --save checkpoints/gpt2_345m_ds --load checkpoints/gpt2_345m_ds --data-path /mnt/ssd-cluster/data/enron/enron_text_document --vocab-file /mnt/ssd-cluster/data/gpt2-vocab.json --merge-file /mnt/ssd-cluster/data/gpt2-merges.txt --data-impl mmap --split 949,50,1 --distributed-backend nccl --lr 3.0e-4 --lr-decay-style cosine --min-lr 3.0e-5 --weight-decay 1e-2 --attention-dropout 0 --hidden-dropout 0 --clip-grad 1.0 --warmup 0.01 --checkpoint-activations --log-interval 1 --save-interval 500 --eval-interval 100 --eval-iters 10 --fp16 --tensorboard-dir tensorboard_data/l_h_1n_8g_2pp_1mp_4b_ds4 --sparsity interspersed --deepspeed --deepspeed_config configs/deepspeed_configs/ds_zero_stage_1_config.json --zero-stage 1 --zero-reduce-bucket-size 50000000 --zero-allgather-bucket-size 5000000000 --zero-contigious-gradients --zero-reduce-scatter --checkpoint-activations --checkpoint-num-layers 1 --partition-activations --checkpoint-in-cpu --synchronize-each-layer --contigious-checkpointing
```

sdtblck:
^ I added a change to get the `megatron/arguments.py` keys programatically, so we don't have to update the list whenever we make changes.

If possible I reckon it would be a good idea to do a similar thing with the deepspeed args, so we don't have to update whenever deepspeed makes changes

StellaAthena:
> 
> 
> This was working and apparently something I did broke it. Currently getting the error
> 
> ```
> deepy.py: error: unrecognized arguments: --model-parallel-size --pipe-parallel-size 2 --num-layers 24 --hidden-size 1024 --num-attention-heads 16 --seq-length 1024 --max-position-embeddings 1024 --batch-size 4 --gas 16 --train-iters 320000 --lr-decay-iters 320000 --save checkpoints/gpt2_345m_ds --load checkpoints/gpt2_345m_ds --data-path /mnt/ssd-cluster/data/enron/enron_text_document --vocab-file /mnt/ssd-cluster/data/gpt2-vocab.json --merge-file /mnt/ssd-cluster/data/gpt2-merges.txt --data-impl mmap --split 949,50,1 --distributed-backend nccl --lr 3.0e-4 --lr-decay-style cosine --min-lr 3.0e-5 --weight-decay 1e-2 --attention-dropout 0 --hidden-dropout 0 --clip-grad 1.0 --warmup 0.01 --checkpoint-activations --log-interval 1 --save-interval 500 --eval-interval 100 --eval-iters 10 --fp16 --tensorboard-dir tensorboard_data/l_h_1n_8g_2pp_1mp_4b_ds4 --sparsity interspersed --deepspeed --deepspeed_config configs/deepspeed_configs/ds_zero_stage_1_config.json --zero-stage 1 --zero-reduce-bucket-size 50000000 --zero-allgather-bucket-size 5000000000 --zero-contigious-gradients --zero-reduce-scatter --checkpoint-activations --checkpoint-num-layers 1 --partition-activations --checkpoint-in-cpu --synchronize-each-layer --contigious-checkpointing
> ```

I fixed the problem

sdtblck:
@StellaAthena what was it?
Also, @joshlk is there a reason for having a fixed set of keys for `ds_config_keys`?
Can we not just feed any key not in `megatron_keys` or `ds_runner_keys` into ds_config_keys, say, in case there's any we're missing, or it changes with updates?

StellaAthena:
> @StellaAthena what was it?
It wasn't the config file, I was calling the function wrong.

> Also, @joshlk is there a reason for having a fixed set of keys for `ds_config_keys`?
> Can we not just feed any key not in `megatron_keys` or `ds_runner_keys` into ds_config_keys, say, in case there's any we're missing, or it changes with updates?
Strongly agree.

**List of things that still need duplication:**
- [ ] batch-size
- [ ] everything in zero-optimization
- [ ] learning rate
- [ ] gradient clipping
- [ ]  deepspeed (should probably be removed)
- [ ] zero-stage
- [ ] zero-reduce-bucket-size
- [ ] zero-allgather-bucket-size
- [ ] zero-contigious-gradients
- [ ] zero-reduce-scatter
- [ ] gradient_clipping

**Other To-Dos:**
- [ ] make the config keys consistent
- [ ] handle arbitrary extra keys
- [ ] organize configs in a way that someone can find what they are looking for

sdtblck:
There are also certain configs which need to be passed in *to both deepspeed and megatron* to work. I.E - zero stage and other zero related settings - which we should make sure we're doing properly

StellaAthena:
> There are also certain configs which need to be passed in _to both deepspeed and megatron_ to work. I.E - zero stage and other zero related settings - which we should make sure we're doing properly

Yeah, that’s why I didn’t just go and delete them. I think ideally we would have the script you wrote handle that, like it handled the batch size. I really don’t want user-facing config files where that has to be listed twice.

joshlk:
I'm still confused about the `train_batch_size` calculation... What I found...

If you start and use 3 nodes and try `ds_pretrain_gpt2_medium_pipe.yml` (`"batch-size":4, "gradient_accumulation_steps":16`) you get the following error...

```
10.141.250.248: AssertionError: Check batch related parameters. train_batch_size is not equal to micro_batch_per_gpu * gradient_acc_step * world_size 1536 != 4 * 16 * 12
```

i.e. its saying the world_size is 12 when there are 24 GPUs. Also if you scroll further up the logs to the pytorch args bit it says `world_size ...................... 24`.

I tried it again with 2 nodes:

```
10.140.23.100:     assert train_batch == micro_batch * grad_acc * self.world_size, \
10.140.23.100: AssertionError: Check batch related parameters. train_batch_size is not equal to micro_batch_per_gpu * gradient_acc_step * world_size 1024 != 4 * 16 * 8
```

So its saying the world size is 8 when there are 16 GPUs and if you scroll up it says `world_size ...................... 16`.

Really quite odd. Ive tested it with 1, 2 & 3 nodes and now works with the "divide by 2" calculation. But if I try another confiuration file `ds_pretrain_gpt2.yml` with 3 nodes (24 GPUs) it makes no sense:

```
10.140.23.78: AssertionError: Check batch related parameters. train_batch_size is not equal to micro_batch_per_gpu * gradient_acc_step * world_size 48 != 4 * 1 * 6
```

i.e. world_size==6 when GPUs == 24?!

---

> For some reason even though lr prints as 0.0003 when args are printed, but during training I get this

> ^ I added a change to get the megatron/arguments.py keys programatically, so we don't have to update the list whenever we make changes.

👍 . We need to be careful about duplicate keys though

> Also, joshlk is there a reason for having a fixed set of keys for ds_config_keys? Can we not just feed any key not in megatron_keys or ds_runner_keys into ds_config_keys, say, in case there's any we're missing, or it changes with updates?

I think its a good idea to [validate the user input](https://github.com/EleutherAI/gpt-neox/blob/ea2f01c81565ca15e90bd219d370bd030c919175/megatron/config_monster.py#L121) and ensure they haven't included any keys that aren't recognised. otherwise this could create strange behaviours. For example, say if the user misspells a megatron param it would then get passed to DS conf instead of megatron and DS will probably just ignore it without the users knowledge.

> - [ ] make the config keys consistent

I think we should use the same parameter names as deepspeed and megatron as much as possible (even though they are inconsistent). This will aid debugging and makes it easier to look up corresponding documentation.

> - [ ] handle arbitrary extra keys

As said above this should and currently does produce an error

joshlk:
@StellaAthena I'm not sure what you mean by these being duplicates, can you be more specific?:

 zero-stage
 zero-reduce-bucket-size
 everything in zero-optimization
 zero-allgather-bucket-size
 zero-contigious-gradients
 zero-reduce-scatter
 gradient_clipping

sdtblck:
@joshlk I know what she means, I'm sorting it :)

sdtblck:
Okay, deduplicated the remaining params.

Zero parameters should be set deepspeed style, like so:
```
   "zero_optimization": {
    "stage": 0,
    "allgather_partitions": True,
    "allgather_bucket_size": 500000000,
    "overlap_comm": False,
    "reduce_scatter": True,
    "reduce_bucket_size": 500000000,
    "contiguous_gradients": False,
    "cpu_offload": False
},
```

same with optimizer params, like so:
```   "optimizer": {
     "type": "Adam",
     "params": {
       "lr": 0.00015,
       "max_grad_norm": 1.0,
       "betas": [0.9, 0.95]
     }
   }
```
(options are "adam", "onebitadam", "cpu_adam", "cpu_torch_adam")

gradient clipping should be set with "gradient_clipping" instead of clip-grads

and i think that's about it.

StellaAthena:
@sdtblck looks a lot better! Do you think there are other params that would be worth bundling together, deepspeed-style? I mostly have the checkpointing args in mind here, I think.

sdtblck:
@StellaAthena i think whether we do checkpointing args deepspeed style or not is inconsequential, really. But i can set it up that way if you think it'd be more user friendly.

sdtblck (approved):

sdtblck (approved):

PR #139, opened 2021-02-27
Fix sparsity

sdtblck:
The sparse attention mask needed to be a float, but was in fact a bool mask, and so was leaking information from future tokens. This fixes that.


sdtblck:
This apparently includes all the commits from `config_monster`, so maybe let's merge that first

PR #147, opened 2021-02-28
Fix different positional embeddings clashing

sdtblck:
As mentioned on discord, if using rpe, the regular pos embs were still being added in. 

This should fix that - you now specify the type of positional embedding you want as the "--pos-emb" arg and can choose from 'learned', 'sinusoidal' and 'rpe'.

Still needs testing.


megatron/model/gpt2_model.py
```diff
@@ -171,17 +165,16 @@ def __init__(self, num_tokentypes=0, parallel_output=True, topology=None):
...
-                                self.init_method,
-                                self.num_tokentypes,
-                                args.sinusoidal_pos_emb))
+                                        self.hidden_size,
+                                        args.padded_vocab_size,
+                                        args.max_position_embeddings,
+                                        args.hidden_dropout,
+                                        self.init_method,
+                                        self.num_tokentypes))
+
```
MicPie, line 177: Is here a `args.pos_emb == "sinusoidal"` missing instead of the previous `args.sinusoidal_pos_emb`?

megatron/model/gpt2_model.py
```diff
@@ -227,20 +220,19 @@ def _logits_helper(embedding, lm_output):
...
-                            tied_weight_attr='word_embeddings_weight')
+                              EmbeddingPipe,
+                              self.hidden_size,
+                              args.padded_vocab_size,
+                              args.max_position_embeddings,
+                              args.hidden_dropout,
+                              self.init_method,
+                              self.num_tokentypes,
+                              forward_fn=_logits_helper,
+                              tied_weight_attr='word_embeddings_weight')
```
MicPie, line 235: Is here a `args.pos_emb == "sinusoidal"` missing instead of the previous `args.sinusoidal_pos_emb`?

megatron/model/gpt2_model.py
sdtblck: nope, i removed it and instead we use get_args() in the embedding layer  https://github.com/EleutherAI/gpt-neox/pull/147/files/64d0683612f65be7b443e9362a5df34d0d93e92b

sdtblck:
We might want to double check our sinusoidal implementation, but all the others apppear to work fine.

image

dashed peach line:  sinusoidal, 
other peach line : learned embedding summed w input, 
green : no pos emb
grey : rpe

StellaAthena:
Are we going to leave this open until the bug in sinusoidal is fixed or would you rather we merged and opened a bug issue @sdtblck

sdtblck:
> Are we going to leave this open until the bug in sinusoidal is fixed or would you rather we merged and opened a bug issue @sdtblck

It's unclear whether it's a bug but if it is, it's in the main branch too. And the rpe stuff just flat out doesn't work in the main branch, so i'd suggest we merge this.

StellaAthena (approved):
This mostly worlds, however it appears that there’s a bug in the sinusoidal embedding.

PR #149, opened 2021-03-01
Cleanup rpe emb. and values.

MicPie:
Cleanup of the rpe variable names for the rpa flag to `rpe`/`self.rpe` and the rpe embedding to `rpe_emb`/`self.rpe_emb`.


sdtblck:
I think our rpe implementation as is won't work correctly, so there's not too much point in merging this, as it's covered in https://github.com/EleutherAI/gpt-neox/pull/147

PR #154, opened 2021-03-03
Add Sm3 optimizer

sdtblck:
@ClashLuke reports sm3 works well training Jannet, so now we have it


StellaAthena:
This PR seems to contain a bunch of commits that are on `main` already. I’m going to fast forward it before merging to clarify the commit history.

joshlk (approved):

PR #156, opened 2021-03-04
Log to file

joshlk:
Save logs to file. For Eluether cluster default location is `/mnt/ssd-cluster/logs`

Not everything is captured as the `deepspeed.launcher` output isn't captured but its a start. I have tested with and without logs turned on. Ready to merge.


sdtblck (approved):
awesome!

PR #157, opened 2021-03-04
Change license to Apache v2

joshlk:
Letting this sit for a bit, so people can think about it.

To two main benefits are:
A) It would allow nvidia/Megatron to merge our changes
B) Having two licences is unnecessarily complicated


deepy.py
```diff
@@ -1,6 +1,6 @@
-#!/usr/bin/env python
-
-# Copyright 2021 (c) Josh Levy-Kramer <josh@levykramer.co.uk>. All rights reserved.
+# coding=utf-8
+# Copyright (c) 2021 EleutherAI
+# This file is based on code by the authors denoted below and has been modified from its original version.
```
joshlk, line 3: A couple of comments:
- we need the shebang as this file is executable. this allows you to run this file like this `./deepy.py` without the need to do `python deepy.py` as it tells the shell interpreter which program to use to run the file
- I don't know why megatron have include `coding=utf-8` as its now obsolete. All python 3 files are utf8 unless otherwise specified
- This files has not been modified from megatron but is a "neox original". So no need for the "This file is based on code by the authors denoted..." preamble

joshlk:
@StellaAthena I've standardised the headers. If you happy lets merge

StellaAthena (approved):

PR #158, opened 2021-03-05
Checkpoint save location and disk space

joshlk:
@sid @StellaAthena 

Suggested save locations for checkpoints. I have also increased disk space so more checkpoints can be saved


joshlk:
@sdtblck @StellaAthena when you kill a cluster, can you use `bash kubernetes/kill_k8s.sh foo` as this will also delete the mount drive associated with it. Otherwise we might start to have loads of dangling mounts that aren't being used

sdtblck (approved):

sdtblck:
lgtm :rocket:

PR #159, opened 2021-03-05
Delete old checkpoints

sdtblck:
adds an arg to keep a rolling window of the last n checkpoints and delete old ones

`keep-last-n-checkpoint` -> defaults to None


sdtblck:
contains commits from https://github.com/EleutherAI/gpt-neox/pull/154 so let's merge that first

joshlk:
Ive added in a default into our configs of keep the last 4

joshlk (approved):

StellaAthena:
Having this default to `None` feels strange to me. Doesn’t that mean we are secretly disabling checkpointing? I think it would make more sense to have it default to a small number (1? 2?) rather than None.

joshlk:
`None` just means that no checkpoints will be deleted - which I think is a sensible default

StellaAthena:
> `None` just means that no checkpoints will be deleted - which I think is a sensible default

Oh okay, I'm totally on board then. I thought it meant none we saved.

PR #162, opened 2021-03-06
HOTFIX BUG: circular import

joshlk:
Fixes:

```
  File "./deepy.py", line 25, in <module>
    from megatron.config_monster import ConfigMonster
  File "/home/mchorse/gpt-neox/megatron/config_monster.py", line 10, in <module>
    from megatron.utils import obtain_resource_pool
  File "/home/mchorse/gpt-neox/megatron/utils.py", line 34, in <module>
    from megatron.checkpointing import save_checkpoint
  File "/home/mchorse/gpt-neox/megatron/checkpointing.py", line 35, in <module>
    from megatron.utils import natural_sort
ImportError: cannot import name 'natural_sort' from partially initialized module 'megatron.utils' (most likely due to a circular import) (/h
```

### DIFF
--- a/megatron/utils.py
+++ b/megatron/utils.py
@@ -31,7 +31,6 @@ from megatron import get_args
 from megatron import print_rank_0
 from megatron import get_adlr_autoresume
 from megatron import mpu
-from megatron.checkpointing import save_checkpoint
 from megatron.data.samplers import DistributedBatchSampler
 from megatron.fp16 import FP16_Optimizer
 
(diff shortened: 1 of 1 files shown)


StellaAthena:
Is this actually going to fix the circular import? By moving the import inside the function call doesn't that delay rather than prevent the problem?

joshlk:
No it solves the issue. It’s the standard way to prevent circula imports and I’ve tested it.

PR #163, opened 2021-03-06
Add model sizes from GPT3 paper + other minor changes

sdtblck:


StellaAthena (approved):
Looking good! A100s go brrrrr

PR #164, opened 2021-03-07
add scalenorm

sdtblck:


ontocord:
Hi team, can you explain why you init the scalenorm variable to 1.0? 

https://github.com/EleutherAI/gpt-neox/blob/43f6330ac953715370bf4cb723e2950f135c5e46/megatron/model/norms.py#L77

In the author's implementation, they set it to args.embed_dim ** 0.5. See https://github.com/tnq177/transformers_without_tears/blob/25026061979916afb193274438f7097945acf9bc/layers.py#L157

PR #166, opened 2021-03-07
Cuda11.1

joshlk:


joshlk:
This is now working. I have tested ssh. I didnt test deepspeed

sdtblck:
tested with a run and everything appears to work fine, merging now

PR #169, opened 2021-03-07
Add some comments and get rid of dead code

leogao2:


StellaAthena (changes requested):
Please update from main and address the conflicts.

StellaAthena:
Leo, I'm confused by your "get rid of dead code" changes. I started to revert it, but realized asking is probably a better idea than acting unilaterally lol. Why do you think we don't use `megatron/model/distributed.py`? It gets called by the main training loop!

StellaAthena (changes requested):
Leo, I'm confused by your "get rid of dead code" changes. I started to revert it, but realized asking is probably a better idea than acting unilaterally lol. Why do you think we don't use `megatron/model/distributed.py`? It gets called by the main training loop! If that file is really dead code, then there's a significant amount of other dead code to remove as well.

Regardless, we cannot merge this into main as-is because it doesn't run. It immediately errors out with `ModuleNotFoundError: No module named 'megatron.model.distributed'`. If you want to remove `megatron/model/distributed.py` you need to make corresponding changes elsewhere so that the code can run.

leogao2:
Can someone please review

sdtblck (approved):

PR #174, opened 2021-03-10
Weight decay groups

sdtblck:
this ensures that there's only one parameter group when weight_decay = 0 (there were only 2 to separate the parameters that should be affected by weight decay and that shouldn't).

This should reduce the amount of calls to `.compressed_allreduce` in onebitadam and hopefully improve efficiency.

also some minor improvements to pipeline parallel logging


StellaAthena (approved):
Looks good to me

PR #175, opened 2021-03-12
Sparse attn fix for mp > 1

sdtblck:
pretty sure this should fix it, but unfortunately sparse attn doesn't work with cuda 11.1 anyway, so i have nowhere I can easily test.


sdtblck:
I just tested on some V100s and this is working for both mp > 1, pp > 1, and (mp > 1 & pp > 1)

StellaAthena (approved):

PR #177, opened 2021-03-12
Bump deeperspeed version. Fix sudo bug

joshlk:
Bump deeperspeed version. Fix sudo bug


StellaAthena (dismissed):
Looks good to me! Note that this DeeperSpeed commit doesn't include @sdtblck's Sparse Attention fix. Do we want to include that?

sdtblck (approved):

sdtblck (approved):

PR #195, opened 2021-03-17
Overhaul of readme + config readme, plus other bits

sdtblck:


StellaAthena (approved):
After a bunch of discussion in discord, people have decided they like it. We may revise further but everyone is happier with this than the current readme

PR #196, opened 2021-03-18
Add pipe-partition-method arg

sdtblck:
Adds an argument which allows you to control the method used to distribute model layers across pipeline stages. 

Choose from "parameters", which balances the number of parameters on each pipeline stage, "uniform", which naively balances the number of layers per stage, or "type:[regex]" (in our instance this will basically only be type:transformer), which balances layers whose class names match [regex]

ignore commit message, it's old lol

### DIFF
--- a/megatron/arguments.py
+++ b/megatron/arguments.py
@@ -435,6 +435,12 @@ def _add_distributed_args(parser):
                        help='Size of the model parallel.')
     group.add_argument('--pipe-parallel-size', type=int, default=0,
                        help='Size of the pipeline parallel. Disable with 0.')
+    group.add_argument('--pipe-partition-method', type=str, default='type:transformer',
+                       help='method used to distribute model layers across pipeline stages. Choose from "parameters", '
+                            'which balances the number of parameters on each pipeline stage, "uniform", which naively '
+                            'balances the number of layers per stage, or "type:[regex]" (in our instance this will '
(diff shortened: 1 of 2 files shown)


StellaAthena:
The doc string tells you to use `type:[regex]` but uses the example of `type:transformer` which isn’t a regex.

sdtblck:
> The doc string tells you to use `type:[regex]` but uses the example of `type:transformer` which isn’t a regex.

what? any string can be a regexp

```python
>>> import re
>>> class1 = "TransformerLayer"
>>> class2 = "MLP"
>>> pattern = "transformer"
>>> bool(re.search(pattern, class1.lower()))
True
>>> bool(re.search(pattern, class2.lower()))
False
```

StellaAthena:
> > The doc string tells you to use `type:[regex]` but uses the example of `type:transformer` which isn’t a regex.
> 
> what? any string can be a regexp
> 
> ```python
> >>> import re
> >>> class1 = "TransformerLayer"
> >>> class2 = "MLP"
> >>> pattern = "transformer"
> >>> bool(re.search(pattern, class1.lower()))
> True
> >>> bool(re.search(pattern, class2.lower()))
> False
> ```

Wow, nevermind. I can’t believe I never knew that.

StellaAthena (approved):

PR #199, opened 2021-03-25
ZeRO-3 Goes Brrrrr

### LINKED ISSUE #171: Take ZeRO 3 for a test drive
**Is your feature request related to a problem? Please describe.**
Model too smol

**Describe the solution you'd like**
DeepSpeed ZeRO-3 is finally public! Let’s take it for a test drive (remember to turn pipeline parallelism off) and see if we can get it to run.

**Describe alternatives you've considered**
Use Pipeline Parallelism

**Additional context**
It probably doesn’t work or needs to be modded to work because DeepSpeed :works-internally:

StellaAthena:
Two things we may want to address before merging:
1. Consistent naming policy for ZeRO config files
2. Some combos of configs will not work. For example, CPU offload requires cpu_adam and either ZeRO 1 or ZeRO 3 but not ZeRO 2. Should we check these requirements in config_monster?

Closes #171


configs/meg4B.yml
```diff
@@ -0,0 +1,88 @@
+# GPT-2 pretraining setup
```
sdtblck, line 1: could you get rid of all these benchmarking configs from the PR? They weren't meant to be merged into the main branch

configs/nccl_compressed_allreduce_test.py
```diff
@@ -0,0 +1,196 @@
+'''
```
sdtblck, line 1: not sure what this is doing here lol, this appears to be a deepspeed unit test?

configs/small_compressed_allreduce_test.yml
```diff
@@ -0,0 +1,82 @@
+# GPT-2 pretraining setup
```
sdtblck, line 1: wat

megatron/config_monster.py
```diff
@@ -174,43 +174,44 @@ def _batch_assertion(world_size, train_batch, micro_batch, grad_acc):
...
-            grad_acc is not None:
+    if train_batch is not None and micro_batch is not None and grad_acc is not None:
+        print("fml")
         return train_batch, micro_batch, grad_acc
 
     # gradient_accumulation_steps needs to be set
-    elif train_batch is not None and \
-            micro_batch is not None:
+    elif train_batch is not None and micro_batch is not None:
+        print("4432423")
```
sdtblck, line 183: ... no words lmao

megatron/model/language_model.py
```diff
@@ -150,7 +151,9 @@ def __init__(self,
                 max_sequence_length, self.hidden_size)
             self._position_embeddings_key = 'position_embeddings'
             # Initialize the position embeddings.
-            self.init_method(self.position_embeddings.weight)
+            with deepspeed.zero.GatheredParameters(self.position_embeddings.weight, modifier_rank=0):
```
sdtblck, line 154: does this context manager interfere with things if we don't have ZeRO-3 on? do we need to add a flag?

sdtblck (commented):
Should we add a warning / raise an error somewhere if someone tries to run ZeRO-3 with pipeline parallel? As it is, it looks like it may just silently run, but not actually use ZeRO-3?

Aside from that there's lots of stray print statements / unecessary files, but once those are cleaned up and my comments are addressed, everything else looks good to me :rocket:

megatron/model/language_model.py
StellaAthena: Nope! It doesn’t interfere with things, even if we have ZeRO completely off

StellaAthena:
> Should we add a warning / raise an error somewhere if someone tries to run ZeRO-3 with pipeline parallel? As it is, it looks like it may just silently run, but not actually use ZeRO-3?

That's a good idea, I'll add that in later today.

> Aside from that there's lots of stray print statements / unecessary files, but once those are cleaned up and my comments are addressed, everything else looks good to me 🚀

Took care of this.

StellaAthena:
@sdtblck okay, it’s ready for review now.

StellaAthena:
@sdtblck Adam parameters updated and conflicts fixed.

StellaAthena:
This PR is on hold again, pending updates to ensure consistency with #249

StellaAthena:
Closing because of divergent codebases: at this point it makes the most sense to redo the integration from scratch.

PR #224, opened 2021-04-10
Rewrite data download and tokenization pipeline

leogao2:
Now supports:

- Lmd based archive handling (this includes efficient zstd and tar handling)
- No more unpacking compressed files
- Multi-file support
- Vastly simplified dataset class

### DIFF
--- a/tools/corpora.py
+++ b/tools/corpora.py
@@ -74,14 +74,8 @@ class DataDownloader(ABC):
 
     @property
     @abstractmethod
-    def filetype(self):
-        """filetype of dataset"""
-        pass
-
-    @property
-    @abstractmethod
-    def url(self):
-        """URL from which to download dataset"""
+    def urls(self):
+        """URLs from which to download dataset"""
         pass
 
     @property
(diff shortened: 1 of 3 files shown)


tools/corpora.py
```diff
@@ -99,44 +93,24 @@ def vocab_file(self):
...
     def exists(self):
         """Checks if the dataset is present"""
         return os.path.isdir(f"{self.base_dir}/{self.name}")
 
     def download(self):
         """downloads dataset"""
-        os.makedirs(self.base_dir, exist_ok=True)
-        os.system(f"wget {self.url} -O {os.path.join(self.base_dir, os.path.basename(self.url))}")
+        for url in self.urls:
+            os.makedirs(self.base_dir, exist_ok=True)
```
sdtblck, line 103: this can be moved outside the for loop

sdtblck (commented):
lgtm! I'll test it out then merge

sdtblck:
I'm getting this error when processing 'PileSubset' - I'm guessing it's encountering an empty document or something?

```python
multiprocessing.pool.RemoteTraceback: 
"""
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "tools/preprocess_data.py", line 98, in encode
    doc_ids[-1].append(Encoder.tokenizer.eod)
IndexError: list index out of range
"""
```

tools/corpora.py
leogao2: Fixed

leogao2:
Added a filter for empty documents

sdtblck:
Ok great, I'm going to add a few more pile datasets before we merge. Can lmd to tar.gz?

leogao2:
It should be able to, and if there's any format it can't do that we need I'll update lmd to support it

sdtblck:
Also, do we want to have the urls point to eaidata, or the eye?

sdtblck (approved):

PR #228, opened 2021-04-12
Fix Adam Defaults + Add Adamw

sdtblck:


StellaAthena:
The `DeepSpeedCPUAdam` optimizer has an AdamW option. It defaults to `true` and can be provided as an argument to the constructor so it should be consistent with this feature. I'll have to double check #199 for compatibility before merging it though.

sdtblck:
Cool, i'll take a look at that and try to add it. The changes to the configs are important though, we had the default adam betas set to 0.95 when they should be 0.999

kingoflolz (approved):
LGTM

PR #231, opened 2021-04-14
Fix wandb running locally

sdtblck:
You should be able to set the environment variable `WANDB_LOCAL=1` before running your script if you're using a local wandb server


StellaAthena:
@sdtblck I have attempted to resolve the conflicts with `main`, take a second look and let me know if there's anything you disagree with.

StellaAthena (approved):
Finally getting this in!

PR #233, opened 2021-04-15
Cache RPE Bucket in rpe

sdtblck:


megatron/mpu/layers.py
```diff
@@ -211,6 +210,11 @@ def __init__(self, causal=True, num_buckets=32, max_distance=128, heads=8, init_
                 device=torch.cuda.current_device(), dtype=args.params_dtype))
             _initialize_affine_weight_gpu(self.weight, init_method,
                                           partition_dim=1, stride=1)
+        self._q_len_cached = None
+        self._k_len_cached = None
+        self._q_pos_cached = None
+        self._k_pos_cached = None
```
kingoflolz, line 216: this doesn't do anything right?

sdtblck: correct lol, holdover. let me remove it.

PR #236, opened 2021-04-15
Add Adafactor

antocapp:
Added arguments for Adafactor optimizer in `arguments.py` as well as the optimizer itself in `training.py`.

### DIFF
--- a/megatron/arguments.py
+++ b/megatron/arguments.py
@@ -370,9 +370,6 @@ def _add_learning_rate_args(parser):
                             'rate, maximum number of iterations, and decay style '
                             'from checkpoint and ignore input arguments.')
 
-    group.add_argument('--use-checkpoint-lr-scheduler', action='store_true',
-                       help='')
-
     # Adafactor
     group.add_argument('--adafactor-decay', type=float, default=-0.8, help='Coefficient used to compute running averages of square')
     group.add_argument('--relative-step', action='store_true', help='If True, time-dependent learning rate is computed instead of external learning rate')
--- a/megatron/arguments.py
+++ b/megatron/arguments.py
@@ -236,6 +236,10 @@ def _add_regularization_args(parser):
                        help='Term added to the denominator to improve'
                             'numerical stability')
     group.add_argument('--momentum', type=float, default=0.0, help='momentum term for sm3 optimizer')
+    group.add_argument('--adafactor-eps1', type=float, default=1e-30, help='Regularization constant for square gradient')
+    group.add_argument('--adafactor-eps2', type=float, default=1e-3, help='Regularization constant for parameter scale')
+    group.add_argument('--adafactor-clip', type=float, default=1.0, help='Threshold of root mean square of final gradient update')
+    group.add_argument('--adafactor-beta1', type=float, default=None, help='Coefficient used for computing running averages of gradient')
 
     return parser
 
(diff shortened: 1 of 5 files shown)


megatron/arguments.py
```diff
@@ -365,6 +370,13 @@ def _add_learning_rate_args(parser):
                             'rate, maximum number of iterations, and decay style '
                             'from checkpoint and ignore input arguments.')
 
+    # Adafactor
+    group.add_argument('--adafactor-decay', type=float, default=-0.8, help='Coefficient used to compute running averages of square')
+    group.add_argument('--relative-step', action='store_true', help='If True, time-dependent learning rate is computed instead of external learning rate')
+    group.add_argument('--scale-parameter', action='store_true', help='If True, learning rate is scaled by root mean square')
+    group.add_argument('--adafactor-warmup', action='store_true', help='Time-dependent learning rate computation depends on whether warm-up initialization is being used')
```
StellaAthena, line 377: I would structure the help-string like the others: `If True, ...."

StellaAthena (dismissed):
Looks good to me.

sdtblck:
@StellaAthena We still need to make the changes in `megatron/config_monster.py`

megatron/arguments.py
antocapp: @StellaAthena what do you think about completely delete one of the parameters `relative_step` and `warmup_init`? I can change the code of Adafactor accordingly, in order to use just one parameter (as you noticed in the Discord channel, one of them is useless), and then add the modified optimizer in `optimizers.py`? I think `relative_step` only would be ok.

antocapp:
> @StellaAthena We still need to make the changes in `megatron/config_monster.py`

Do these newly added arguments need to be (here)[https://github.com/EleutherAI/gpt-neox/blob/022e79799e63260e635af52086cf8e339f6fa1b2/megatron/config_monster.py#L90]?

StellaAthena:
> > @StellaAthena We still need to make the changes in `megatron/config_monster.py`
> 
> Do these newly added arguments need to be [here](https://github.com/EleutherAI/gpt-neox/blob/022e79799e63260e635af52086cf8e339f6fa1b2/megatron/config_monster.py#L90)?

Yeah. Also `Adafactor` needs to be added to the list of optimizers and you need to write a function to check that parameter combinations are valid, if there are combos of params that are not.

sdtblck:
I think this should work - give me a few minutes to test

sdtblck:
Ok so this works for me, but I can't get any results other than overflow haha. Can anyone else reproduce? Maybe Adafactor isn't really compatible with fp16?

antocapp:
@sdtblck could it be the `@torch.no_grad()` decorator missing from the step function of the optimizer?

antocapp:
@sdtblck @StellaAthena is it possibile to wrap the Adafactor optimizer inside this class: https://github.com/EleutherAI/gpt-neox/blob/dacaf57e21e623dab4349e51e773fc2d4ae6068c/megatron/fp16/fp16.py#L90 ?

sdtblck:
Hey @antocapp , that should be happening in deepspeed. I'm not sure if the FP16 optimizer in deepspeed is the same as in fairseq, however

StellaAthena:
Can you modify one of the config files to use Adafactor or create a new `small_adafactor.yml` to demonstrate it's features? In my test Adafactor got stuck in a gradient overflow loop and it never came out, but I'm unsure if that's primarily due to me doing it incorrectly or a problem with the code.

StellaAthena (changes requested):
This branch needs to be updated to ensure consistency with #249.

sdtblck:
> This branch needs to be updated to ensure consistency with #249.

@antocapp I'd be happy to do the changes if you can get adafactor working at all on the old branch

StellaAthena:
Closing due to inactivity.

PR #241, opened 2021-04-18
Rotary fixes

sdtblck:
This should bring rotary up to almost the same speeds as learned embeddings, remove permutations and fix rotary + pipeline parallel, which wasn't working previously.

Also some mild reorganisation + added rotary pos emb 'base' as a configurable argument.

tests on a small model:
rotary old = pink, rotary new = purple, learned = orange
Screenshot from 2021-04-18 22-34-15
Screenshot from 2021-04-18 22-34-03

### DIFF
--- a/megatron/model/positional_embeddings.py
+++ b/megatron/model/positional_embeddings.py
@@ -4,9 +4,9 @@ from megatron.module import MegatronModule
 
 class SinusoidalPositionalEmbedding(MegatronModule):
 
-    def __init__(self, dim, theta=10000):
+    def __init__(self, dim, base=10000):
         super().__init__()
-        inv_freq = 1. / (theta ** (torch.arange(0, dim, 2).float() / dim))
+        inv_freq = 1. / (base ** (torch.arange(0, dim, 2).float() / dim))
         self.register_buffer('inv_freq', inv_freq)
 
     def forward(self, x, seq_dim=1):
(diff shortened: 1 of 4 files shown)


megatron/model/positional_embeddings.py
```diff
@@ -0,0 +1,50 @@
...
+            freqs = torch.einsum('i,j->ij', t, self.inv_freq)
+            emb = torch.cat((freqs, freqs), dim=-1).to(x.device)
+            self.cos_cached = emb.cos()[:, None, None, :]
+            self.sin_cached = emb.sin()[:, None, None, :]
+        return self.cos_cached, self.sin_cached
+
+
+# rotary pos emb helpers:
+
+@torch.jit.script
```
kingoflolz, line 43: I don't believe you need a jit in this, as you have a jit on the apply_rotary_pos_emb already

kingoflolz (approved):
LGTM

megatron/model/positional_embeddings.py
sdtblck: ah yeah, that was left on from when i was trying to narrow down the jit bug

kingoflolz (approved):

megatron/model/positional_embeddings.py
```diff
@@ -0,0 +1,49 @@
...
+            self.cos_cached = emb.cos()[:, None, None, :]
+            self.sin_cached = emb.sin()[:, None, None, :]
+        return self.cos_cached, self.sin_cached
+
+
+# rotary pos emb helpers:
+
+def rotate_half(x):
+    x1, x2 = x[..., :x.shape[-1] // 2], x[..., x.shape[-1] // 2:]
+    return torch.cat((-x2, x1), dim=x1.ndim - 1) # dim=-1 triggers a bug in earlier torch versions
```
irasin, line 45: Hi, @sdtblck @kingoflolz @lucidrains  
I have a question about the rotate_half
why change it from rotate_every_two to rotate_half here?

To do rotary embedding, we need to transform the q from [q0, q1, q2, q3, ...] to [-q1, q0, -q3, q2, ... ], right?

PR #251, opened 2021-04-25
fix checkpointing bug

sdtblck:
This bug was so fucking dumb lol.

To give a brief explainer, checkpoint loading was only working correctly if you'd saved + loaded a checkpoint previously.

Spent a good few days trying to figure out why, and it turns out in 'load_checkpoint' (not save_checkpoint, of course) megatron set a global variable called `checkpoint version`. (https://github.com/EleutherAI/gpt-neox/blob/main/megatron/checkpointing.py#L292) 

`checkpoint version` controls how some layers are transposed in the self attention block. (https://github.com/EleutherAI/gpt-neox/blob/main/megatron/model/transformer.py#L300) But since it's only set on load, we *never set this global variable until we load a checkpoint* and therefore after the very first loading from a checkpoint, the way the last dimension is transposed in the transformer layer is totally different, and it fucks everything up.

Extremely stupid, backwards compatibility is a fuck. Megatron code big dumb.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,7 @@ wandb/
 
 # data files 
 data/
+
+# ckpt files
+*.pt
+*.ckpt
--- a/megatron/checkpointing.py
+++ b/megatron/checkpointing.py
@@ -34,21 +34,6 @@ from megatron import get_args
 from megatron import print_rank_0
 from megatron.utils import natural_sort
 
-_CHECKPOINT_VERSION = None
-
-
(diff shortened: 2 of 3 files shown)


sdtblck:
Grey is the first checkpoint load before we fix the bug, blue is the first checkpoint load after

Screenshot from 2021-04-25 14-26-50

ShivanshuPurohit (approved):

PR #253, opened 2021-04-26
Fix recomputed sampling bugs during unconditional text generation

alexandonian:
First, the `forward_model` function now returns a tuple containing `(logits, layer_past)`, but this was not accounted for during the `recompute` mode of operation.

Second, an indentation error caused the `forward_model` function to be called twice when using `recompute` mode. The logits indexing line must also be indented otherwise the highest scoring logits end up being incorrectly sampled from the batch dimension.


sdtblck (approved):
LGTM, thanks Alex! :rocket:

PR #257, opened 2021-04-26
Add Gradient Noise Scale measurement / logging

sdtblck:
from `An Empirical Model of Large-Batch Training` - https://arxiv.org/abs/1812.06162
Should enable us to determine an ideal batch size / batch size schedule for training.

Requires you update your DeeperSpeed to commit hash https://github.com/EleutherAI/DeeperSpeed/commit/750f2140bf782cffeb578ce14a4e4cdb076f4326 or later.

Add `"log_gradient_noise_scale": true` to your config to test

### DIFF
--- a/megatron/gradient_noise_scale/gradient_noise_scale.py
+++ b/megatron/gradient_noise_scale/gradient_noise_scale.py
@@ -48,13 +48,14 @@ class GradientNoiseScale:
         self.n_batches = n_batches
         self.beta = beta
         self.model = model
-        self.model.store_gradients = True
         self.buffer = []
         self.ema_scale = None
         self.ema_noise = None
         self.noise_scale = None
         self.n_updates = 0
         self.cpu_offload = cpu_offload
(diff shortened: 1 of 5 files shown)


megatron/gradient_noise_scale/gradient_noise_scale.py
```diff
@@ -0,0 +1,159 @@
...
+            overflow = is_overflow
+        return overflow
+
+    def _update(self):
+
+        grad = self.flatten_grads()
+        is_overflow = self._sync_overflow(grad is None)
+        if is_overflow:
+            return
+        self.buffer.append(grad)
```
kingoflolz, line 95: Isn't this super expensive? Why not do += and then a division

sdtblck: Ah yep well spotted. Updated version works much better. Thanks!

kingoflolz (approved):
LGTM

PR #259, opened 2021-04-27
Added enwik8 dataset

StellaAthena:


sdtblck:
LGTM :rocket: 
I also added the char level tokenizer. I could only get the preprocessing working without multiprocessing - i think there is maybe some lock on ZipFiles being read by multiple processes. But it takes like 10 seconds to tokenize anyway so I don't think it's a problem :laughing:

sdtblck (approved):

PR #261, opened 2021-04-27
Fix num_workers in preprocessing for enwik8 to 1

sdtblck:


kingoflolz (approved):
LGTM, what's the reasoning for this?

PR #262, opened 2021-04-27
remove --split-sentence from preprocess_data.py

pacman82:
A little bit less code.


sdtblck:
Awesome! We can even slim it down further by removing all the unused nltk stuff + the requirement.

sdtblck (approved):

PR #267, opened 2021-04-27
[BUGFIX] fix final linear layer

sdtblck:
So I did a big woops and, since the final linear layer when weight tying was turned off didn't inherit from torch.nn.module, its parameters weren't even getting registered :facepalm: 

This should fix that.


kingoflolz (approved):
LGTM, what problems did not registering the parameters cause?

sdtblck:
> LGTM, what problems did not registering the parameters cause?

didn't actually see any problems caused by it but it meant we failed some unit tests lol - better safe than sorry I guess

PR #270, opened 2021-04-28
some comments on indexed_dataset.py

pacman82:
Just some comments for someone who wants to figure out, what the memory mapped file format looks like on disk.


StellaAthena (approved):
Love this kind of content!

PR #277, opened 2021-04-30
Remove global vars

sweinbach:
This pull requests removes the global variables. Exception are mpu global variables. These are more complicated and potentially breaking due to being used in torch. autograd.Function implementations.

Train for some steps on small config results in equal loss graphs:

image

Tests are adjusted to the new setup and pass:

image


StellaAthena (approved):
This is amazing! Great job.

megatron/text_generation_utils.py
```diff
@@ -239,24 +230,24 @@ def sample_sequence_batch(model, context_tokens, context_lengths,
...
-                    positions2use, # position_ids
-                    attention_mask, # attention_mask
-                    layer_past, # layer_past
-                )
-                logits, layer_past = forward_model(model, model_inputs)
-                # TODO: we are replicating computation across all machines here, which is really unecessary, we should probably just do it on one then communicate the results
-                logits = logits[:, -1].view(batch_size, -1).contiguous()
-
-            if args.greedy:
+            # we have to use neox_args instead of kwargs here because deepspeed :|
```
sdtblck, line 233: This reverts this bugfix here https://github.com/EleutherAI/gpt-neox/pull/253

megatron/utils.py
```diff
@@ -167,8 +165,11 @@ def get_ltor_masks_and_position_ids(data,
 
 def local_rank():
     """ Local rank of process """
-    return int(os.environ["LOCAL_RANK"])
-
+    local_rank = os.environ.get("LOCAL_RANK")
+    if local_rank is None:
+        print("utils.local_rank() environment variable LOCAL_RANK not set, defaulting to 0", flush=True)
```
sdtblck, line 170: I think we should default to -1 here? (torch / deepspeed's default if local rank isn't set, i.e we're not distributed.)

sweinbach: Some functions rely on rank being equal to 0 (e.g. tensorboard initialization https://github.com/EleutherAI/gpt-neox/blob/319ad174693badc7b1ffde3a2b4b3d068dda588c/megatron/neox_arguments/arguments.py#L106). Would we need to change these functions according to the print_rank_0 function (https://github.com/EleutherAI/gpt-neox/blob/319ad174693badc7b1ffde3a2b4b3d068dda588c/megatron/__init__.py#L17)?

PR #284, opened 2021-05-01
Refactor how we configure sparsity 

sdtblck:
- adds 'attention_config' param which inherits from neo

```
    """
    Attention configuration for gpt-neox
    
    The first item in the list specifies the attention type(s), and should be a list of strings. The second item 
    specifies the number of times to repeat those attention types in the full list.
    
    attention type choices:  [global, local (`sparse_fixed` with no global tokens), sparse_fixed, sparse_variable]
                                
    So a 12 layer network with only global attention could be specified like:
        [[[`global`], 12]]
    
    or a 12 layer network with alternating global / local like:
        [[[`global`, `local`], 6]]
    
    You can set a pattern to propagate across all layers like:
        [[[`global`, `local`], "all"]]

    If no attention_config is specified, this defaults to
        [[[`global`], n_layers]]

    """
```
- adds a `sparsity_config` param - exactly the same as the way deepspeed defines it [here]( https://www.deepspeed.ai/docs/config-json/#sparse-attention) with some minor differences

```
    """
    Sparsity configuration dict as defined in https://www.deepspeed.ai/docs/config-json/#sparse-attention
    
    Note that since neox is autoregressive, attention is always "unidirectional" and `horizontal_global_attention` is 
    always false.
    
    The main difference between our sparsity config and deepspeed's is that `mode` is ignored - since it is instead 
    specified in attention_config defining each layer.
    
    An example config is given below:
          "sparse_attention": {
            "block": 16,
            "different_layout_per_head": true,
            "num_local_blocks": 4,
            "num_global_blocks": 1,
            "num_different_global_patterns": 4,
            "num_random_blocks": 0,
            "local_window_blocks": [4],
            "global_block_indices": [0],
            "global_block_end_indices": None,
            "num_sliding_window_blocks": 3
          }
    """
```

Just adding `sparse.yml` to your deepy command should be the same as setting "sparsity": "interspersed" before.
i.e `./deepy.py pretrain_gpt2.py -d configs small local_setup sparse` 

adding `local` to the attention config is identical to adding `sparse_variable` and setting num_global_blocks to 0.

For `local` to work you need to update deeperspeed to git+git://github.com/EleutherAI/DeeperSpeed.git@311de8b82456da81998c14f33a07d73436cf11be#egg=deepspeed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,4 @@ test_checkpoint/
 test_logs/
 logs/
 tensorboard/
+src/
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ You can also optionally pass in a config prefix, which will assume all your conf
 Example usage:
 
 ```bash
-./deepy.py pretrain_gpt2.py -d configs small_1.yml local_setup.yml
+./deepy.py pretrain_gpt2.py -d configs small.yml local_setup.yml
 ```
 
 This will deploy the `pretrain_gpt2.py` script on all nodes with one process per GPU. The worker nodes and number of GPUs are specified in the `/job/hostfile` file (see [parameter documentation](configs)), or can simply be passed in as the `num_gpus` arg if running on a single node setup.
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ You can also optionally pass in a config prefix, which will assume all your conf
 Example usage:
 
 ```bash
-./deepy.py pretrain_gpt2.py -d configs small.yml local_setup.yml
+./deepy.py pretrain_gpt2.py -d configs small_1.yml local_setup.yml
 ```
 
 This will deploy the `pretrain_gpt2.py` script on all nodes with one process per GPU. The worker nodes and number of GPUs are specified in the `/job/hostfile` file (see [parameter documentation](configs)), or can simply be passed in as the `num_gpus` arg if running on a single node setup.
--- a/configs/sparse.yml
+++ b/configs/sparse.yml
@@ -1,0 +1,13 @@
+# Add this to your config for sparse attention every other layer
+{
+  "attention_config": [[["global", "sparse_variable"], "all"]],
+  # sparse attention config (these are the defaults, training will work without this here, but it's left in for
+  # illustrative purposes)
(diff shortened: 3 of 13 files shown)


sdtblck:
Added sliding window attention too - will need to update to deeperspeed 73dfe6b1ae6209a4195defcc271866986337e95b for it to work

sweinbach:
LocalSlidingWindowSparsityConfig did not import. Reason is that it is not imported here: https://github.com/EleutherAI/DeeperSpeed/blob/73dfe6b1ae6209a4195defcc271866986337e95b/deepspeed/ops/sparse_attention/__init__.py#L1

Fixed the import for now in neox: https://github.com/EleutherAI/gpt-neox/pull/284/commits/79cd3c1d9956e27dc375488f89dffe4a00eba460

sweinbach:
Added a testcase for model instantiation with the sparsity config. This testcase is not passing. Seems like the implementation in DeeperSpeed needs some rework. 


  File "/home/aleph/repos/neox/gpt-neox-copy/src/deepspeed/deepspeed/ops/sparse_attention/sparsity_config.py", line 687, in __init__
    super().__init__(num_heads, block, different_layout_per_head)
NameError: name 'different_layout_per_head' is not defined

sdtblck:
ahh thanks @sweinbach I hadn't actually tested it yet, let me get that working properly now!

sweinbach:
> ahh thanks @sweinbach I hadn't actually tested it yet, let me get that working properly now!

Shall we also add testcases for other configurations?

sdtblck:
> > ahh thanks @sweinbach I hadn't actually tested it yet, let me get that working properly now!
> 
> Shall we also add testcases for other configurations?

I think test cases should parametrized to cover as many configurations as possible, yeah! I haven't really had a chance to look at the tests yet though.

sdtblck:
Ok @sweinbach it should work now with deeperspeed version 353387d794b4083f4333b717b90862cb3d6bc53b

ShivanshuPurohit (approved):

StellaAthena (approved):

PR #287, opened 2021-05-02
fix: update storage class + fix duplicate name?

jamiecore:


kubernetes/k8_spec_ssd-cluster.yml
```diff
@@ -1,13 +1,13 @@
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
-  name: eleuther-ssd-0
+  name: eleuther-ssd-cluster
```
StellaAthena, line 4: `ssd-0` is how we have been labeling pod-independent permanent storage. `ssd-cluster` is the designation for the storage associated with the particular cluster that you are currently on.

jamiecore: ok just noticed the same `metadata.name` in both the `k8_spec_ssd-cluster.yml` and `k8_spec_ssd-0.yml`

StellaAthena: Hmmm. Looking through it, it does seem strange. But I just ran this a couple days ago without issue, and that makes me worried about changing it.

sdtblck:
Hi @jamiecore - we're now using https://github.com/EleutherAI/kube for our deployment, the folder here will be deleted soon

PR #298, opened 2021-05-05
Fix growing memory usage in `preprocess_data.py`

sdtblck:
Memory usage in preprocess_data.py was insane, because `_multi_lmd` was processing much faster than the actual tokenization, and imap was storing up the results in memory.

Adding a semaphore object causes multi_lmd to wait until the tokenizing processes needs more data and fixes the memory problem.


kingoflolz (approved):
Looks good, but does this really require a triton and deepspeed version bump?

sdtblck:
> Looks good, but does this really require a triton and deepspeed version bump?

Version bumps are in main already - triton took  triton==1.0.0.dev20210502 off of pip, 0.4.0 is actually the same version

PR #299, opened 2021-05-05
Generation bugfix + refacturing

sweinbach:
Started with fixing the batch size bug in the first commit and ended up with the refacturing of text generation. 

main changes:
- Fix batching bug in inference
- Functions now pass through generation parameters.
- removal of dead code
- Getting interactive mode to run

I kept the generation config in NeoXArgs in order to pass from deepy to text_gen_gpt2.py.


sweinbach:
thanks for the review @sdtblck 

- rename max_tokens to maximum tokens -> done
- defaults to old values -> done
- rephrase neox_args docstring -> done
- use tokenizer.eod instead of tokenizer.eod_id -> done
I changed it to make transparent that the token id (int) not the token (string) is used. This is how huggingface implements it, which lead to my confusion. Looking at the not implemented property in the abstract class, this makes sense though.
- user script setting -> done
- reformat long code lines in text generation utils -> done
- fix recompute -> done
The logic is the same (apart from the now fixed error of forgetting to reset layer_past for recompute). There is one change: recompute now only forwards all tokens up to the position in question. It used to be all tokens (padded to full seq_length). This compute is not necessary / later tokens are never used.

I wanted to remove duplicate lines (forward_model, logic for recompute and first iteration the same, get rid of counter variable only used here but spread through the function). I find the version now easier to understand. Could be the bias of having it written though.
 
- done logic to a separate function -> still thinking about it, let's discuss it

sdtblck (approved):

PR #307, opened 2021-05-07
scale d_ff for equal parameter count with geglu + fix geglu bias

sdtblck:
Fixes geglu bias to be in accordance with https://arxiv.org/pdf/2002.05202.pdf (GELU(xW+b)⊗(xV+c))

and scales down the ff dims by 2/3 so we don't need to worry about parametrizing d_ff or balancing gelu vs. geglu experiments


kingoflolz (approved):
LGTM, if you use this you gotta be careful that ff_dim is actually sensible and can run efficiently (as it will not be a nice multiple of your d_ff)

PR #308, opened 2021-05-08
Fix gas checkpoint resuming

alexandonian:
Despite best efforts to fix checkpoint resuming, we've identified an additional bug that occurs when resuming a run with `gradient_accumulation_steps` set to a value above one. In particular, a training run that required continual restarts due to job time limits began producing the following loss trajectory:

gas_bug

Clearly, the run is not resuming in the exact same state as it left off, and small scale experiments isolated the issue the gas > 1 setting. The loss curves suggests that the model was revisiting data it had just previously encountered before the checkpoint. Indeed, upon resuming, the dataloader start iteration was not being fast forwarded in a manner that took into account the gradient accumulation steps, and this PR addresses this issue. This update appears to have restored the step-for-step perfect checkpoint resuming as seen below:

gas_fix

### DIFF
--- a/megatron/training.py
+++ b/megatron/training.py
@@ -36,6 +36,7 @@ from megatron import mpu
 from megatron.model import GPT2ModelPipe
 from megatron.checkpointing import load_checkpoint, save_checkpoint
 from megatron.data.gpt2_dataset import build_train_valid_test_datasets
+from megatron.data.gpt2_dataset import build_the_dataset
 
 from megatron.initialize import initialize_megatron
 from megatron.learning_rates import AnnealingLR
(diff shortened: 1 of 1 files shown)


kingoflolz (dismissed):
Looks good, could you remove the whitespace/formatting changes though?

sdtblck:
Also we need to

- Make the same changes for validation dataloader resuming
- Make sure non pipe parallel training is actually respecting g.a.s (Currently i don't think it is). Would be good to make all these changes in one PR.

sdtblck:
Ok, I made the changes above. Think it should be good to merge now.

sdtblck (approved):

PR #316, opened 2021-05-10
Add alternative init methods

ConnorJL:
This PR adds two new model parameters, "init_method" and "output_layer_init_method", which set the init method of all non-residual and the residual FF-outputs (I think), respectively. The options are "normal", "scaled_normal" (GPT style), "orthogonal", "xavier_uniform" and "xavier_normal". I've tested at least that they all run for a few hundred steps on a small model. Worth noting that orthogonal init is *extremely* slow on first init.

### DIFF
--- a/megatron/model/init_functions.py
+++ b/megatron/model/init_functions.py
@@ -79,6 +79,8 @@ def xavier_normal_init_method():
     return init_
 
 def small_init_init_method(dim):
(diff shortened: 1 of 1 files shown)


sdtblck:
Can we move the init methods to a separate file? I feel like utils is not really the right place.

 I would also like to add SmallInit / WangInit as we discussed in the discord.

Aside from that, lgtm.

sdtblck (approved):

PR #321, opened 2021-05-11
Add blended datasets

sdtblck:
Adds the ability to feed in multiple datasets + blend them with weights (see examples in local_setup.yml / eleutherai_cluster.yml

### DIFF
--- a/megatron/data/data_utils.py
+++ b/megatron/data/data_utils.py
@@ -199,6 +199,7 @@ def weights_by_num_docs(l, alpha=0.3):
 
     Hence α (`alpha`) allows us to control how much to 'boost' the probability of training on low-resource groups.
 
+    See https://arxiv.org/abs/1911.02116 for more details
     """
     total_n_docs = sum(l)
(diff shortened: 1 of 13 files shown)


sdtblck:
+ adds the ability to weight the datasets by number of documents as in https://arxiv.org/abs/1911.02116

megatron/data/data_utils.py
```diff
@@ -0,0 +1,295 @@
...
+                            int(round(split * float(size))))
+    diff = splits_index[-1] - size
+    for index in range(1, len(splits_index)):
+        splits_index[index] -= diff
+    assert len(splits_index) == 4
+    assert splits_index[-1] == size
+    return splits_index
+
+
+def get_normalized_weights_and_num_samples(data_paths: List[str], weights: List[float], num_samples: int) -> Tuple[List[float], List[int]]:
```
sweinbach, line 135: We could remove the data_paths arg as it is not actually used.

PR #323, opened 2021-05-12
feat(model): add patched orthogonal init

ClashLuke:
In the tradition introduced by kindiana, I named this patch after myself.


sdtblck:
I've changed the name to scaled orthogonal since it's more informative :laughing: Wang Init would have been scaled init if it wasn't already taken

sdtblck (approved):

PR #329, opened 2021-05-14
save original configs with checkpoint

sweinbach:
Problem: neox args could not be instantiated from configs.yml saved with the checkpoint. Reason is the exponential notation which got saved as e.g. "1e10" (instead of "1.0e10"). On load of yaml this is interpreted as string. 

Solution: copy the original config files to the checkpoint directory. This way also potential comments are not lost. Files are copied upon instantiation of neox args in deepy.py

Assumption (unchanged): only of version of config is saved. A potential history of changed configs during training is not available in the checkpoint directory.

### DIFF
--- a/deepy.py
+++ b/deepy.py
@@ -17,9 +17,7 @@ import logging
 import os
 
 import deepspeed
-import requests
 from deepspeed.launcher.runner import main
-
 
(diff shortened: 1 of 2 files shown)


sdtblck:
LGTM :rocket:

sdtblck (approved):

PR #327, opened 2021-05-14
Readme update

StellaAthena:

PR #335, opened 2021-05-18
fix(readme): add missing contributors

ClashLuke:


ClashLuke:
image
@sdtblck figured out why you could edit the branch in the old pull request. apparently, there's a flag you can set to opt-out of the base maintainers pushing to the fork.

StellaAthena:
#327 is an already existing pull request that adds additional contributors as well as many other things. I’m going to close this as duplicative, but double check that the two lists are the same and push to that branch if there is someone missing.

PR #341, opened 2021-05-19
make sure data_dir arg gets passed on to vocab / merge file etc

sdtblck:


StellaAthena:
What is the failure case this aims to correct?

sdtblck:
If you pass in your own data dir as an argument, the tokenizer vocab and merge file still tries to download to the default data dir

StellaAthena (approved):

Issue #366, opened 2021-06-23
Eval Harness doesn't log during training


StellaAthena:
The [eval harness](www.github.com/eleutherai/lm-eval-harness) has been integrated (#312) but currently can only be run separately. It would be nice to be able to set eval tasks and log the results during training.

PR #367, opened 2021-06-24
Run eval harness during training

sdtblck:
addresses https://github.com/EleutherAI/gpt-neox/issues/366

The generation tasks aren't that fast - so best to stick to the log likelihood tasks (you can add them to the yaml with the "eval_tasks" parameter.

e.g

```yaml
"eval_tasks": ["lambada", "wikitext", "piqa"],
```

### DIFF
--- a/eval_tasks/__init__.py
+++ b/eval_tasks/__init__.py
@@ -1,0 +1,1 @@
+from .adaptor import EvalHarnessAdaptor, run_eval_harness
--- a/eval_tasks/adaptor.py
+++ b/eval_tasks/adaptor.py
@@ -1,18 +1,21 @@
-from megatron.utils import is_local_main 
+from megatron.utils import is_local_main
 import best_download
 
 # patch best_download (eval harness downloader) to only happen on the first rank
 fn = best_download.download_file
 
+
 def _download_file(*args, **kwargs):
     if is_local_main():
         fn(*args, **kwargs)
+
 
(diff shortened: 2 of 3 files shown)


StellaAthena (changes requested):
It looks like it's locking on piqa. We need to check out the HF dataset builder internals and make sure it's thread safe.

StellaAthena:
__Notes on this PR's behavior:__
1. running just lambada works fine
2. running just piqa works fine

StellaAthena:
mathqa + piqa has been running for over 8 hours without any problems. I’m wondering if the core problem is about mixing HF and non-HF tasks

sdtblck:
Hey @StellaAthena can you test out the above change, and let me know if it makes a difference?
I try to pre-download all the tasks on local rank 0, to get rid of any multithreading problems

StellaAthena:
> Hey @StellaAthena can you test out the above change, and let me know if it makes a difference?
> I try to pre-download all the tasks on local rank 0, to get rid of any multithreading problems

It's on today's TODO list :)

sweinbach (approved):

sdtblck:
Just some peace of mind for @StellaAthena that this is definitely working :) 
Screenshot from 2021-08-31 16-36-22
Screenshot from 2021-08-31 16-38-07

StellaAthena (approved):

PR #378, opened 2021-07-19
[temporary] fix to index errors in datasets

sdtblck:


StellaAthena (changes requested):
This is a very temporary patch, but it's good to be able to run the code.

StellaAthena (approved):
Clicked "request changes" by accident :/

PR #383, opened 2021-07-30
Use lru_cache for GPT2Tokenizer.bpe

iczero:
GPT2Tokenizer currently uses an unbounded cache (`GPT2Tokenizer.cache`) which will grow to extremely large sizes (a few million entries according to objgraph) when running `tools/preprocss_data.py`, leading to massive memory use, especially with many workers. 

Performance appears to be affected slightly.


sdtblck:
Haven't personally experienced the memory leak myself, but the PR lgtm so I'll merge now.

Thanks!

sdtblck (approved):

Issue #377, opened 2021-07-19
Index Errors with Weighted Dataset


sdtblck:
**Describe the bug**
When training with the blended datasets, at some point in the validation, we get an index error trying to load a data item

**Proposed solution**
We can just try except and return a random index if we get an index error, but I'd prefer to eventually know the root cause for the problem

**Screenshots**
![image](https://user-images.githubusercontent.com/46172032/126128381-9cba6909-2ec3-4130-a0ce-6648ad1afc46.png)

sdtblck:
closed by https://github.com/EleutherAI/gpt-neox/pull/384

PR #384, opened 2021-07-30
Fix bug with number of evaluation steps

sdtblck:
we were running way too many evaluation steps if the model is pipe parallel + has g.a.s on because of this line

```python
            for _ in range(neox_args.gradient_accumulation_steps):
```

fixing this to 1 if the model is pipe parallel fixes the issue, as .eval_batch() already takes gradient accumulation steps into account.

```python
for _ in range(1 if neox_args.is_pipe_parallel else neox_args.gradient_accumulation_steps):
```

This also relates to [this issue](https://github.com/EleutherAI/gpt-neox/issues/377), and is probably the reason we were seeing a stopiteration error


sweinbach (approved):

sweinbach:
Tested in a running training as well...

PR #404, opened 2021-09-06
Correct gradient accumulation

hyunwoongko:
1. Accumulated gradient must be updated at once. If you update every step, the effect of accumulation is lost.
2. We need to divide the loss collected in the list into a gradient accumulation steps. Currently, you are dividing by world size.

### DIFF
--- a/megatron/training.py
+++ b/megatron/training.py
@@ -447,26 +447,28 @@ def train_step(neox_args, timers, data_iterator, model, optimizer, lr_scheduler)
             )
             timers("forward").stop()
             losses.append(loss)
-            # Calculate gradients, reduce across processes, and clip.
-            timers("backward").start()
-            backward_step(
+        
+        # Calculate gradients, reduce by gradient accumulation steps, and clip.
+        reduced_loss = torch.cat([loss.clone().detach().view(1) for loss in losses]).mean()
+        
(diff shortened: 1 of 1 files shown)


megatron/training.py
```diff
@@ -457,16 +456,19 @@ def train_step(neox_args, timers, data_iterator, model, optimizer, lr_scheduler)
...
-                raise ValueError("Must be using deepspeed to run neox")
-            timers("optimizer").stop()
-        reduced_loss = {
-            "lm_loss": reduce_losses(losses).mean()
-        }  # reduces losses across machines for logging
+                
+        # Update parameters.
+        timers("optimizer").start()
+        if neox_args.deepspeed:
+            model.step()
```
hyunwoongko, line 463: - https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L1463
- https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L1414

hyunwoongko: https://medium.com/huggingface/training-larger-batches-practical-tips-on-1-gpu-multi-gpu-distributed-setups-ec88c3e51255

hyunwoongko:
I will reupload after more checking.

PR #414, opened 2021-09-28
fix launcher args being passed in and resource inclusion

sdtblck:
The deepspeed launcher args weren't being passed in in the right position. (launcher args should come before user script / args)

Also, using the --include / --exclude flags wasn't working properly (you can't specify those and num_gpus at the same time, weird stuff was happening with batch size calculations...)

### DIFF
--- a/deepy.py
+++ b/deepy.py
@@ -27,11 +27,6 @@ from megatron.utils import get_wandb_api_key
 
 
 neox_args = NeoXArgs.consume_deepy_args()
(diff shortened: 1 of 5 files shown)


sdtblck:
Not quite ready to merge yet, one thing to fix...

sdtblck:
Should be ready to merge now. Also updated requirements

EricHallahan (approved):

Issue #419, opened 2021-10-11
Running through Dockerfile broken


VHellendoorn:
**Describe the bug**
When using an image based on the provided `Dockerfile` and running the quick start steps (download enron data, run `deep.py`), execution crashes before training begins.

**To Reproduce**
1. Build an image using the provided Dockerfile
2. Run said image, mounting 8 RTX800 GPUs
3. Fetch enron data using the `prepare_dataset.py` script
4. Run `./deepy.py pretrain_gpt2.py -d configs small.yml local_configs.yml`
5. The code crashes with a non-descript `NCCL error in: /pytorch/torch/lib/c10d/ProcessGroupNCCL.cpp:825, unhandled system error, NCCL version 2.7.8`

**Expected behavior**
Training starts or a specific error is provided.

**Proposed solution**
The NCCL error is typically a stand-in for a real issue that is not relayed back through multiprocessing. As a first step, it would be nice to know if this setup works out-of-the-box for others; in that case, it might be my resources or CUDA version.

**Environment (please complete the following information):**
 - GPUs: 8 RTX8000 GPUs
- Configs: Ubuntu 20.04, Cuda 11.2

**Additional context**

StellaAthena:
Based on [this issue](https://github.com/PyTorchLightning/pytorch-lightning/issues/4420) I would try downgrading to pytorch 1.6 and cuda 10.2. It seems that the problem is GPU-specific and I do not have access to a RTX8000 for testing unfortunately.

VHellendoorn:
Thanks for your quick responses. I did dig into that linked issue thread a bit, but no luck with the fixes proposed there.

@EricHallahan, I was referring to the output of `nvcc --version` within the container, which on a second look is 11.1 (not 11.2), my apologies. The specific build is `cuda_11.1.TC455_06.29190527_0`, so I'm guessing driver version 455.06. This is all run through an [nvidia-docker backed container](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html) (version 2.6.0-1) on Linux; the driver/CUDA version in the host OS is 450.142 / 11.0, but I suspect that discrepancy doesn't play a role. No changes were made to the code and nothing else was installed in the container.

The container did prompt me to run that `python` command the first time I ran `deep.py`, and installing that seemed to have been successful. So the full chain of commands is:

```bash
git clone https://github.com/EleutherAI/gpt-neox && cd gpt-neox
[...]

PR #424, opened 2021-10-15
Add Docker instructions to README

VHellendoorn:
Provides the instructions for using the `Dockerfile` to run the code, as discussed in #419. Feedback welcome; hope this is helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,8 +51,9 @@ If you're looking for our TPU codebase, see [GPT-Neo](https://github.com/Eleuthe
 
 Coming soon: a colab notebook for trying out the model.
 
-**Warning:** Our codebase relies on [DeeperSpeed](https://github.com/EleutherAI/DeeperSpeed), our fork of the [DeepSpeed](https://github.com/microsoft/DeepSpeed) library with some added changes. We strongly recommend using Anaconda, a virtual machine, or some other form of environment isolation before installing from `requirements/requirements.txt`. Failure to do so may cause other repositories that rely on DeepSpeed to break.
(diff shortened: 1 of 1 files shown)


VHellendoorn:
Absolutely! I'm currently training a models at around the 1B parameter mark (e.g., from `large.yml`) on roughly 150GB of data on a single node with 8 RTX8000 GPUs. At that scale, I'm seeing about 45 TFLOPS per GPU, which translates into ~18 sequences of 2,048 tokens per second on the full node. With better memory utilization, I can push the latter to about 25. That about echoes the realistic maximum these cards output and near-ideal node-level parallelism (about 7x the per-GPU performance); utilization is at a steady 100%.

I am planning to venture into the few-billion parameter territory after this, but realistically I need more than one server to train at that level. In my own benchmarking, I can get a throughput of about 7 sequences per second on the 2-7B model, so (with gradient-accumulation to accomodate larger batch sizes) that means training about 3x slower (makes sense; it's 3x the parameters). I'm already not getting through much more than 5GB of data per day, so training a larger model once on 150GB would take many months.

In terms of configuration, I tried a number of sweeps on the `large` configuration to calibrate. The results is that I am not currently using the model or pipeline parallelism -- the former slowed training down dramatically (maybe because no NVLink) and the latter was about even, though worse for larger batch sizes (probably because this is all intra-node). I do always enable the two flags for softmax and gelu fusion; they tend to add neglible setup time and boosted performance by a pretty significant amount in a few ad-hoc experiments. The stage-1 ZeRO optimizations seem to help as well; higher stages don't really seem to pay off, but maybe they will yet. And 1-bit Adam had no discernible impact whatsoever, which again makes some sense for single-node training. Oh and sparse training did speed things up a good bit, but led to errors when using in text-generation mode, some of which I could fix with padding and others that seemed more fundamental. I'm deferring on that one for a while.

I think that captures all my experiments. Hope this helps!

PR #453, opened 2021-11-03
[WIP] Add Adafactor Optimizer

Quentin-Anthony:
Mostly adapted from: https://huggingface.co/transformers/_modules/transformers/optimization.html

Preliminary training results on 4 V100 GPUs within a node for ~300 steps show increased GPU memory utilization (https://wandb.ai/quentin-anthony/neox?workspace=user-quentin-anthony), but slightly higher samples/sec. 

adafactor-samples
adafactor-gpumem

I'm using the adafactor2-7B.yml config vs. standard 2-7B.yml

### DIFF
--- a/configs/adafactor2-7B.yml
+++ b/configs/adafactor2-7B.yml
@@ -1,0 +1,82 @@
+# GPT-2 pretraining setup
+{
+   # parallelism settings ( you will want to change these based on your cluster setup, ideally scheduling pipeline stages
+   # across the node boundaries )
+   "pipe-parallel-size": 1,
+   "model-parallel-size": 1,
+
+   # model settings
+   "num-layers": 32,
+   "hidden-size": 2560,
+   "num-attention-heads": 32,
+   "seq-length": 2048,
+   "max-position-embeddings": 2048,
+   "norm": "layernorm",
+   "pos-emb": "rotary",
+   "no-weight-tying": true,
+
+   # these should provide some speedup but takes a while to build, set to true if desired
+   "scaled-upper-triang-masked-softmax-fusion": false,
+   "bias-gelu-fusion": false,
+
+   
+   # optimizer settings
+   "optimizer": {
+     "type": "Adafactor",
+     "params": {
+     }
+   },
+   "zero_optimization": {
+    "stage": 1,
+    "allgather_partitions": True,
+    "allgather_bucket_size": 500000000,
+    "overlap_comm": True,
+    "reduce_scatter": True,
+    "reduce_bucket_size": 500000000,
+    "contiguous_gradients": True,
+    "cpu_offload": False
+  },
+
+   # batch / data settings
+   "train_micro_batch_size_per_gpu": 4,
+   "data-impl": "mmap",
+   "split": "949,50,1",
+
+   # activation checkpointing
+   "checkpoint-activations": true,
+   "checkpoint-num-layers": 1,
+   "partition-activations": true,
+   "synchronize-each-layer": true,
+
+   # regularization
+   "gradient_clipping": 1.0,
+   "weight-decay": 0,
+   "hidden-dropout": 0,
+   "attention-dropout": 0,
+
+   # precision settings
+   "fp16": { 
+     "fp16": true,
+     "enabled": true,
+     "loss_scale": 0,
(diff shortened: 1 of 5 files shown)


StellaAthena:
@Quentin-Anthony Do you have an idea why it’s running slower than the baseline version? I haven't read the Adafactor paper in a while, but is this the expected outcome? I’m not seeing where the benefit of using it might come from.

StellaAthena (changes requested):
The demo config file should explicitly set the optimizer's h parameters. While the code will run as written due to the use of default values, it's much more pedagogically useful to demonstrate the h params in the config file.

Quentin-Anthony:
> The demo config file should explicitly set the optimizer's h parameters. While the code will run as written due to the use of default values, it's much more pedagogically useful to demonstrate the h params in the config file.

Will do

StellaAthena:
> > @Quentin-Anthony Do you have an idea why it’s running slower than the baseline version? I haven't read the Adafactor paper in a while, but is this the expected outcome? I’m not seeing where the benefit of using it might come from.
> 
> We should be seeing less memory usage than Adam, so this is unexpected. I haven't had the time to look into this yet. Should be soon.

Have you had a chance to look at this? I think we shouldn’t merge it until we’re confidant it’s functioning as anticipated.

Quentin-Anthony:
> > > @Quentin-Anthony Do you have an idea why it’s running slower than the baseline version? I haven't read the Adafactor paper in a while, but is this the expected outcome? I’m not seeing where the benefit of using it might come from.
> > 
> > 
> > We should be seeing less memory usage than Adam, so this is unexpected. I haven't had the time to look into this yet. Should be soon.
> 
> Have you had a chance to look at this? I think we shouldn’t merge it until we’re confidant it’s functioning as anticipated.

I've tried several adafactor settings, but no luck. I agree we shouldn't merge anything yet. 

I'll do some digging on how Huggingface sets up adafactor. Maybe we'll get some clues there. In the meantime, I'm open to suggestions.

StellaAthena:
I spoke with user `finetune` in the EAI discord server and he says he found that the HF implementation worked fine out of the box. So hopefully it won’t be too much of a bother.

That said, the recent paper out of DeepMind doesn’t cast Adafactor in a particularly good light. After reading that, it’s unclear to me if we should really care about getting Adafactor merged at all…

Quentin-Anthony:
The latest status here is that I've tried the following primary configurations:

```
class Adafactor(Optimizer):
    def __init__(
        self,
        params,
        lr=None,
        eps=(1e-30, 1e-3),
        clip_threshold=1.0,
        decay_rate=-0.8,
        beta1=None,
        weight_decay=0.0,
        scale_parameter=True,
        relative_step=True,
        warmup_init=False,
    ):

```


<pre>
class Adafactor(Optimizer):
    def __init__(
        self,
        params,
        lr=None,
        eps=(1e-30, 1e-3),
        clip_threshold=1.0,
        decay_rate=-0.8,
        beta1=None,
        weight_decay=0.0,
        <b>scale_parameter=False,</b>
        relative_step=True,
        warmup_init=False,
    ):
</pre>


<pre>
class Adafactor(Optimizer):
    def __init__(
        self,
        params,
        <b>lr=.001,</b>
        eps=(1e-30, 1e-3),
        clip_threshold=1.0,
        decay_rate=-0.8,
        beta1=None,
        weight_decay=0.0,
        <b>scale_parameter=False,</b>
        <b>relative_step=False,</b>
        <b>warmup_init=False,</b>
    ):

</pre>

I've also tried adjusting the LR, batch size, etc. No notable decrease in GPU memory allocated for 2-7B and 6-7B models . This adafactor implementation is functionally equivalent to that of HF. 

Unless I'm missing something obvious, Adafactor doesn't work for us. If anybody else is more familiar with Adafactor, I'm open to suggestions.

Quentin-Anthony:
I would recommend that we either: 

A) Close this PR without merging since we don't see benefits.
B) Merge and keep it optional. Since the code is equivalent to HF and my wandb results show the same loss over time as Adam, I think the implementation itself is correct.

StellaAthena:
A while ago we decided to not support Adafactor due to its performance issues. Closing this issue as we seem to have forgotten to do so at the time.

Issue #459, opened 2021-11-07
Fix checkpoint / config file management


sdtblck:
Currently, every time NeoXArgs are initialized, the old configs directory is deleted, and new configs are saved out. (see https://github.com/EleutherAI/gpt-neox/blob/main/megatron/neox_arguments/arguments.py#L235)

This was done originally because:
    1) we wanted to retain comments / formatting in the configs - doing something like saving neox_args.to_dict() would not do that.
    2) we wanted to retain separation of files (you might have one config controlling model size and another data loading settings, for example)
    3) we wanted to avoid old configs sitting around in the configs directory

But it introduces a few problems:
    1) whenever you load a model for inference, neox will attempt to delete and overwrite the configs. But really we would only want this to happen when we're saving out a new checkpoint.
    2) if you don't have permission to write to the disk where the configs are stored, it will error out.

Proposed solution: 
    - Hold a copy of the original config files in memory, then only save them out when *saving* a checkpoint?

sweinbach:
Good idea. I suggest to write the config files then to the global step directory. This would allow seeing the exact config from which the checkpoint originates and keep track of changes in the config over the course of Training (e.g. when adapting batch size, learning rate etc)

StellaAthena:
Currently we don’t accurate track changing config files mid training. While many h params can’t change during training, it’s conceivable that someone might manually adjust the learning rate. Also, when changing to a different GPU cluster someone might adjust the parallelism settings.

Currently, the code doesn’t have a way to track this. Do we wish to?

sdtblck:
> relatedly, we currently don’t accurate track changing config files mid training. While many h params can’t change during training, it’s conceivable that someone might manually adjust the learning rate. Also, when changing to a different GPU cluster someone might adjust the parallelism settings.

since we save out new configs every time training is run, we do track changing config files mid training. Unless I'm missing something.

sdtblck:
> Good idea. I suggest to write the config files then to the global step directory. This would allow seeing the exact config from which the checkpoint originates and keep track of changes in the config over the course of Training (e.g. when adapting batch size, learning rate etc)

This makes sense to me, yeah

sweinbach:
> > relatedly, we currently don’t accurate track changing config files mid training. While many h params can’t change during training, it’s conceivable that someone might manually adjust the learning rate. Also, when changing to a different GPU cluster someone might adjust the parallelism settings.
> 
[...]

PR #463, opened 2021-11-09
Fix checkpoint / config file management

sweinbach:
Saves config files in the global step directory instead of the neox_args.save directory. Discussion here: https://github.com/EleutherAI/gpt-neox/issues/459

Tested on small config with:
- mp 1, pp 1
- mp 2, pp 1
- mp1, pp 2

Question is how to get the config files from the load in deepy to the single processes. I opted for the hand-trough via neox args. Any other idea?

### DIFF
--- a/megatron/checkpointing.py
+++ b/megatron/checkpointing.py
@@ -152,10 +152,7 @@ def save_ds_checkpoint(iteration, model, neox_args):
         logits = do_forward_pass(neox_args=neox_args, model=model)
         sd['checkpoint_validation_logits'] = logits
     
-    # Derive tag 
-    # The tag is usually automatically derived in the same way in deepspeed
-    # As the save_checkpoint function does not return such derived tag, we 
-    # do it manually in order to use the subdirectory to save config files
+    # checkpoint folder name
     tag = f"global_step{iteration}"
 
     # save checkpoint
--- a/megatron/checkpointing.py
+++ b/megatron/checkpointing.py
@@ -152,7 +152,22 @@ def save_ds_checkpoint(iteration, model, neox_args):
         logits = do_forward_pass(neox_args=neox_args, model=model)
         sd['checkpoint_validation_logits'] = logits
     
-    model.save_checkpoint(neox_args.save, client_state=sd)
+    # Derive tag 
+    # The tag is usually automatically derived in the same way in deepspeed
+    # As the save_checkpoint function does not return such derived tag, we 
+    # do it manually in order to use the subdirectory to save config files
+    tag = f"global_step{iteration}"
+
+    # save checkpoint
+    model.save_checkpoint(neox_args.save, tag=tag, client_state=sd)
+
+    # save config files
+    if  torch.distributed.get_rank() == 0 and neox_args.config_files is not None:
+        configs_directory = os.path.join(neox_args.save, tag, "configs")
+        os.makedirs(configs_directory)
+        for config_filename, config_data in neox_args.config_files.items():
+            with open(os.path.join(configs_directory, config_filename), "w") as f:
+                f.write(config_data)
 
 def save_checkpoint(neox_args, iteration, model, optimizer, lr_scheduler):
     """Save a model checkpoint."""
--- a/megatron/neox_arguments/neox_args.py
+++ b/megatron/neox_arguments/neox_args.py
@@ -709,6 +709,11 @@ class NeoXArgsTraining(NeoXArgsTemplate):
     Output directory to save checkpoints to.
     """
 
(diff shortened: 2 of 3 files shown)


StellaAthena:
It’s probably worth testing with MP > 1, PP > 1, and two machines. That’s about as maximally asynchronous as it’ll be.

megatron/checkpointing.py
```diff
@@ -152,7 +152,22 @@ def save_ds_checkpoint(iteration, model, neox_args):
         logits = do_forward_pass(neox_args=neox_args, model=model)
         sd['checkpoint_validation_logits'] = logits
     
-    model.save_checkpoint(neox_args.save, client_state=sd)
+    # Derive tag 
+    # The tag is usually automatically derived in the same way in deepspeed
+    # As the save_checkpoint function does not return such derived tag, we 
+    # do it manually in order to use the subdirectory to save config files
+    tag = f"global_step{iteration}"
```
sdtblck, line 159: I would say these comments are a bit more verbose than necessary - I think what's happening here is straightforward enough that we can just say something like `# Derive tag` or `# checkpoint folder name` and save a few LOC

megatron/checkpointing.py
```diff
@@ -152,7 +152,22 @@ def save_ds_checkpoint(iteration, model, neox_args):
...
+
+    # save checkpoint
+    model.save_checkpoint(neox_args.save, tag=tag, client_state=sd)
+
+    # save config files
+    if  torch.distributed.get_rank() == 0 and neox_args.config_files is not None:
```
sdtblck, line 165: should this be local_rank == 0? What if there isn't a shared filesystem

megatron/neox_arguments/neox_args.py
```diff
@@ -709,6 +709,11 @@ class NeoXArgsTraining(NeoXArgsTemplate):
     Output directory to save checkpoints to.
     """
 
+    config_files: dict = None
```
sdtblck, line 712: I'm wondering if we should initialize non-configurable variables like this in __post_init__ - I don't think they really warrant a docstring, so it might be a better place for them.

sdtblck (changes requested):
LGTM but I would remove some of the comments for the tag name and just double check the `torch.distributed.get_rank() == 0` thing

sweinbach:
> It’s probably worth testing with MP > 1, PP > 1, and two machines. That’s about as maximally asynchronous as it’ll be.

Can do but probably only tomorrow.

megatron/checkpointing.py
sweinbach: Following reasoning. Hence I would like to keep it only for rank 0.

- also without shared file system the checkpoint is only complete with data from all ranks. There is no added benefit, you would not use only part of the checkpoint.
- duplicating data across the ranks then would require overwrites when moving the files to one location
- I have had at least one shared file system with problems writing to the same file from multiple ranks. This would avoid such multiple write handles. Deepspeed does it as well for latest file and the zero conversion script. Maybe should not be a problem, but it seems to be one some times.

megatron/checkpointing.py
sweinbach: No opinion really. Can change :-)

megatron/neox_arguments/neox_args.py
sweinbach: My thinking was to facilitate development for all not having seen this PR by making all properties of neox args transparent. If some contributor wants to use this info it is nice to read it in the docstring and not be required to read through code to actually see that it exists and what it is.

megatron/checkpointing.py
sweinbach: Deepspeed DOES write to one file at the same time from multiple ranks. To avoid confusion…

megatron/neox_arguments/neox_args.py
sweinbach: Ah also: the way it is implemented the file contents are part of neox args and distributed by the deepspeed launcher. This means it needs to be a named argument in the current implementation. So the property needs to be defined and then probably also be documented.

megatron/checkpointing.py
sweinbach: done

sweinbach:
> > It’s probably worth testing with MP > 1, PP > 1, and two machines. That’s about as maximally asynchronous as it’ll be.
> 
> Can do but probably earliest tomorrow.

Now also tested with pp 2 and mp4. Works.

StellaAthena:
@sdtblck How does this look to you? It looks ready to merge to me.

sdtblck (approved):

PR #464, opened 2021-11-09
Supersede max_position_embeddings by seq_length

Muennighoff:
It seems that `max_position_embeddings` is redundant and we can save one argument by only sticking to `seq_length`


EricHallahan:
Although `max_positional_embeddings` may seem redundant given its strong correlation in the sample configurations with `seq_length`, they have different purposes and semantics. [`max_position_embeddings` sets the number of GPT-style learned positional embeddings to initialize with the model](https://github.com/EleutherAI/gpt-neox/blob/570d10cee66ab3b40a8443e65bb34ecc572a0769/megatron/model/word_embeddings.py#L60-L65) and [represents a model hyperparameter](https://github.com/EleutherAI/gpt-neox/blob/904ea338bd3405452d1b2930743e4cfb3a2f5e38/megatron/checkpointing.py#L130-L142), while `seq_length` represents the sequence length to process during training and inference. Since the sequence dimension of the embedding matrix limits the sequence length when using models with learned absolute positional embeddings, [we enforce that `max_position_embeddings >= seq_length`](https://github.com/EleutherAI/gpt-neox/blob/74afd1b10fdb3b23bd55f0763cb2e61221196363/megatron/neox_arguments/arguments.py#L708-L713).

As for why it is useful to decouple these values, [techniques like the staged sequence length training presented in Shortformer](https://arxiv.org/abs/2012.15832) rely on the separate values for `max_positional_embeddings` and `seq_length`. This also may be useful when fine-tuning models on downstream tasks or limiting the sequence length seen by models at inference. This is not a oversight and is very much an intentional feature of GPT-NeoX.

Muennighoff:
Nice thanks for seeing this, didnt think of shortformer; well we can at least fix the typo in the max_position_embedding explanation if we keep it; https://github.com/EleutherAI/gpt-neox/pull/465

PR #465, opened 2021-11-10
Fix max_position_embedding typo

Muennighoff:


StellaAthena:
Thank you for pointing this out. We are currently rewriting and overhauling the docstrings and will make sure that this typo doesn't occur.

PR #468, opened 2021-11-17
Remove redundant if-else and fix typos

satpalsr:
Modified the print_rank_0 function to follow "If distributed is initialized print only on rank 0."

### DIFF
--- a/megatron/__init__.py
+++ b/megatron/__init__.py
@@ -16,8 +16,12 @@ import torch
 
 def print_rank_0(*message):
     """If distributed is initialized print only on rank 0."""
-    if torch.distributed.is_initialized() and torch.distributed.get_rank() == 0:
+    if torch.distributed.is_initialized():
+        if torch.distributed.get_rank() == 0:
+            print(*message, flush=True)
+    else:
         print(*message, flush=True)
+
 
 from .initialize import initialize_megatron
 from .neox_arguments import NeoXArgs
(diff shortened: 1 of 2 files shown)


EricHallahan (approved):
Looks good to me! I think there are a few other errata like this around the codebase that could use some cleaning, but this certainly is a good place to start.

StellaAthena (changes requested):
Thanks for correcting some typos! Your PR description makes it seem like the change to `megatron/__init__.py` has no functional impact, but that doesn't look like it's the case to me. Can you clarify?

satpalsr:
Hey @StellaAthena , you are right. The function did got changed due to slight misunderstanding.
My commit can be corrected as shown below:
```
def print_rank_0(*message):
    """Print message except when distributed is initialized and rank not 0."""
    if not (torch.distributed.is_initialized() and torch.distributed.get_rank() != 0):
        print(*message, flush=True)
```
Or I can switch back to the original version:
```
def print_rank_0(*message):
    """If distributed is initialized print only on rank 0."""
    if torch.distributed.is_initialized():
        if torch.distributed.get_rank() == 0:
            print(*message, flush=True)
    else:
        print(*message, flush=True)
```
What looks better to you?

StellaAthena:
I think that the second looks better tbh. It’s less elegant, but given the level of complexity of the code base we often compromise on elegance in favor of transparency. I need to stop and think for a second about the line you wrote to understand what it does due to the compound logical expression, and that seems like that’s an unnecessary pitfall for new users. I would prefer to keep that as is.

StellaAthena (approved):

PR #477, opened 2021-12-07
add log_grad_pct_zeros to neox_args

CoEich:
Added functionality to log percentage of zeros in gradients to detect underflows. 

Also added option to initialize wandb on all ranks so all gradient stats get logged when using pipeline parallelism. 

Did I understand the issue correctly? Adding the above change made it work for me.


StellaAthena:
Ooo this is quite useful, thank you for the addition.

Can you post WandB links to runs showing it both in normal operation and when detecting an issue?

CoEich:
So far I don't have any examples with a nontrivial percentage of zero gradients. Still think the addition might be useful in the future, at least it doesn't hurt to have it.

StellaAthena:
> So far I don't have any examples with a nontrivial percentage of zero gradients. Still think the addition might be useful in the future, at least it doesn't hurt to have it.

I don't disagree, my comment is more directed at telling whether it works as expected.

CoEich:
Ok - Could you specify what exactly you would like to see, e.g just an example run where the stat gets logged or that it actually detects underflows? I'd be happy to resolve this. Also thanks for taking the time to look at it :-)

sdtblck:
HeY @CoEich thanks for the PR! I agree this looks pretty helpful. 

One thing we could do to test is to just modify the grads to be very small and see if it logs as expected, but looking at the code, this is all pretty straightforward, and the codepath isn't entered unless you explicitly set this option to true, so i'd say I'm happy to merge if @StellaAthena approves.

sdtblck (dismissed):

StellaAthena:
> Ok - Could you specify what exactly you would like to see, e.g just an example run where the stat gets logged or that it actually detects underflows? I'd be happy to resolve this. Also thanks for taking the time to look at it :-)

Two example runs, one where it detects an underflow and one where an underflow doesn’t occur.

StellaAthena (approved):
This works. We don't have a good guide for usage yet, but leaving the PR here to rot sees worse than just merging it and making docs later.

PR #480, opened 2021-12-15
Model and config code of an HF gpt-neox model; a conversion script.

ZHAOTING:
The modeling and configuration files are largely based on HF's gpt-j model.
(I found gpt-j's architecture more similar to gpt-neox than gpt-neo, especially it uses rotary embedding)

Modifications to the original gpt-j modeling:

- Added post-attention layernorm as `ln_2`.
- Changed `q_proj, k_proj, v_proj` linear layers to a single `qkv_proj` that corresponds to gpt-neox's `attention.query_key_value` linear layer. And set `bias=True`.
- Combined gpt-neox's and HF gpt-j's rotatry embedding functions.
- Set `bias=False` for `lm_head`.
- Updated the computation in `GPTNeoXBlock` in correspondence to two residual computing ways in gpt-neox, which is controlled by a new config argument `gpt_j_residual`.

Modifications to the original gpt-j configuration:

- Set the default value of `activation_function` to `gelu`.
- Removed `rotary_dim` (so that its default value is `None`).
- Added a `gpt_j_residual` argument (default value is `False`) in correspondence to two residual computing ways in gpt-neox.

A conversion script:

- that reads config files and gpt-neox's output state dict files and outputs a pretrained HF pytorch GPTNeoX model. 
- Note that weights are not loaded for two kinds of model parameters `transformer.h.*.attn.bias` and `transformer.h.*.attn.masked_bias` because they should keep their default values.

Things that I have checked with a 1B model, which is trained basically following the default `XL.yml` config:

- The above conversion script works correctly.
- The greedy-decoding outputs by gpt-neox's inference script and HF's `generate()` are identical.
- Intermediate outputs (e.g. hidden states) are *almost* identical when running from gpt-neox code and HF code. There are some small differences, which I think are caused by precision settings.

Things that are not included in this pull request:

- Tensorflow-related code.
- Conversion script that considers checkpoints trained with model parallel.
- Other model variants that might use a different type of, for example, rotary embeddings.
- Things that haven't come to my mind.

BTW, I haven't found a better place to put the new files so I simply created a directory `huggingface` under `/tools`.

A table that summarizes parameters (and their shapes) in 1) a GPT-NeoX checkpoint, 2) an HF GPTNeo model, 3) an HF GPTJ model, and  4) the HF GPTNeoX model in this pull request:
Screen Shot 2021-12-15 at 13 29 32


eghbalhosseini:
which transformers version is this compatible with, gpt_neox requirement is transformer=4.5.0 (from requirement.txt). however the gpt_model in huggingface doesn't have self.post_init() until version 4.13.0

ZHAOTING:
@eghbalhosseini I tested it with a pretty late version, possibly 4.15.0.
If we are to decide whether it should have `self.post_init()` or not, I think it's better to align the model code with HF rather than gpt-neox, because it should be part of HF transformers.

StellaAthena:
I think this is an awesome initiative, though I do worry that deploying our 20B pretrained model with HF will be unviable due to computational inefficiencies in the `transformers` library. This isn’t a knock against HF, but their library is optimized for ease of use for end-users in the 100M - 1B scale and I worry that for > 10B models it will be very hard to use, especially for training.

We do plan on training and releasing smaller models though, and I view this integration as a net good. Just want to temper expectations in case you’re hoping that this will enable people to finetune a 20B model.

what can I do to help?

StellaAthena:
@ZHAOTING That's a lot of config files... so far we have trained a 125M model, a 760M model, 1.3B model, and a 20B model. Did you train a 1B model using this codebase?

ZHAOTING:
@StellaAthena Yes, I have trained a 1.3B model on some different data using this codebase.

eghbalhosseini:
I want to benchmark a small model trained using gpe-neox, and after conversion using HF pipelines, @StellaAthena would it be possible to do it on 125M model ?

StellaAthena:
> I want to benchmark a small model trained using gpe-neox, and after conversion using HF pipelines, @StellaAthena would it be possible to do it on 125M model ?

DM me on Discord and I'll hook you up with one

eghbalhosseini:
@StellaAthena  just did! thanks.

eghbalhosseini:
just a clarifying question @ZHAOTING , If I understand the conversation code, the config and model are not not taking type of positional embedding, layer normalization into account, something the gpt-neox offers. is this true? do we have to pass arguments to initialize models based on normalization and embedding spec?

StellaAthena:
> just a clarifying question @ZHAOTING , If I understand the conversation code, the config and model are not not taking type of positional embedding, layer normalization into account, something the gpt-neox offers. is this true? do we have to pass arguments to initialize models based on normalization and embedding spec?


In the interest of making things as downstream transparent as possible, HF has tried to make each "class" a precise model architecture - Change one thing, and you need a new class.

This comes with a huge burden to developers though, especially as they forbid using inheritance. I  recommend asking for guidance from HuggingFace: https://github.com/huggingface/transformers/issues/15642

ZHAOTING:
@eghbalhosseini 
Yes, the HF code and conversion script do not support position embedding types and layer norm positions other than the default values. 
Therefore, there are no arguments for these features and we don't need to pass such arguments for initialization.

eghbalhosseini:
@ZHAOTING what I am worried about is that the HF model is only written for 1 instance of models one can create in gpt-neox library. so we probably have to extend this so that that it has a compatible HF model with each configuration of gpt-neox

ZHAOTING:
If we are to extend the HF code to be compatible with all model configurations, a big barrier is that we need corresponding pretrained models to verify the HF implementation.

eghbalhosseini:
right, I think the one immediate thing we can do to do that verification to extract hidden states from a model implemented in gpt-neox and HF counterpart. HF has already a pipeline to do this, so what we need is a version for gpt-neox

eghbalhosseini:
I have written a wrapper for learned position embedding model, and did some initial test. now I want to do a comparison of hidden states with GPT-NEOX output, to make sure the models do the same computation. for a benchmark that I ran they are consistent, but want to make sure. any thoughts on how I can extract embeddings from gpt-neox?

0x7o (approved):

oborchers:
@ZHAOTING: Am I the only one where the provided script isn't working anymore? I tried this on the raw checkpoints, as well as on the merged checkpoints from the `mp_merge` branch, but I always run into:

Or is this to be applied on the full checkpoint?

```
RuntimeError: Error(s) in loading state_dict for GPTNeoXForCausalLM:
	size mismatch for transformer.wte.weight: copying a param with shape torch.Size([25216, 6144]) from checkpoint, the shape in current model is torch.Size([50400, 6144]).
....
	size mismatch for lm_head.weight: copying a param with shape torch.Size([25216, 6144]) from checkpoint, the shape in current model is torch.Size([50400, 6144]).
```

lopho:
> @ZHAOTING: Am I the only one where the provided script isn't working anymore? I tried this on the raw checkpoints, as well as on the merged checkpoints from the `mp_merge` branch, but I always run into:
> 
> Or is this to be applied on the full checkpoint?
> 
> ```
> RuntimeError: Error(s) in loading state_dict for GPTNeoXForCausalLM:
> 	size mismatch for transformer.wte.weight: copying a param with shape torch.Size([25216, 6144]) from checkpoint, the shape in current model is torch.Size([50400, 6144]).
> ....
> 	size mismatch for lm_head.weight: copying a param with shape torch.Size([25216, 6144]) from checkpoint, the shape in current model is torch.Size([50400, 6144]).
> ```

got a similar error running generation on the vanilla branch (1xA6000, ```  "pipe-parallel-size": 1,
  "model-parallel-size": 1,```).

```
[2022-03-30 19:30:19,967] [INFO] [engine.py:84:__init__] CONFIG: micro_batches=32 micro_batch_size=4
[2022-03-30 19:30:19,986] [INFO] [engine.py:141:__init__] RANK=0 STAGE=0 LAYERS=49 [0, 49) STAGE_PARAMS=20552994816 (20552.995M) TOTAL_PARAMS=20552994816 (20552.995M) UNIQUE_PARAMS=20552994816 (20552.995M)
 > number of parameters on model parallel rank 0: 20552994816
 > total params: 20,552,994,816
[2022-03-30 19:30:20,003] [INFO] [engine.py:1551:_load_checkpoint] rank: 0 loading checkpoint: ./20B_checkpoints/global_step150000/mp_rank_00_model_states.pt
  File "generate.py", line 88, in <module>
    main()
  File "generate.py", line 32, in main
    model, neox_args = setup_for_inference_or_eval(use_cache=True)
  File "/media/zweistecken/work/ART/gpt-neox-20b/gpt-neox/megatron/utils.py", line 433, in setup_for_inference_or_eval
    model, _, _ = setup_model_and_optimizer(
  File "/media/zweistecken/work/ART/gpt-neox-20b/gpt-neox/megatron/training.py", line 436, in setup_model_and_optimizer
    neox_args.iteration = load_checkpoint(
  File "/media/zweistecken/work/ART/gpt-neox-20b/gpt-neox/megatron/checkpointing.py", line 233, in load_checkpoint
    checkpoint_name, state_dict = model.load_checkpoint(
  File "/media/zweistecken/work/ART/gpt-neox-20b/env/lib/python3.8/site-packages/deepspeed/runtime/engine.py", line 1523, in load_checkpoint
    load_path, client_states = self._load_checkpoint(load_dir,
  File "/media/zweistecken/work/ART/gpt-neox-20b/env/lib/python3.8/site-packages/deepspeed/runtime/engine.py", line 1558, in _load_checkpoint
    self.load_module_state_dict(state_dict=checkpoint['module'],
  File "/media/zweistecken/work/ART/gpt-neox-20b/env/lib/python3.8/site-packages/deepspeed/runtime/pipe/engine.py", line 1278, in load_module_state_dict
    self.module.load_state_dir(load_dir=self._curr_ckpt_path, strict=strict)
  File "/media/zweistecken/work/ART/gpt-neox-20b/env/lib/python3.8/site-packages/deepspeed/runtime/pipe/module.py", line 571, in load_state_dir
    layer.load_state_dict(torch.load(model_ckpt_path,
  File "/media/zweistecken/work/ART/gpt-neox-20b/env/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1497, in load_state_dict
    raise RuntimeError('Error(s) in loading state_dict for {}:\n\t{}'.format(
RuntimeError: Error(s) in loading state_dict for EmbeddingPipe:
	size mismatch for word_embeddings.weight: copying a param with shape torch.Size([25216, 6144]) from checkpoint, the shape in current model is torch.Size([50304, 6144]).

````

PR #500, opened 2022-02-02
Update Makefile

StellaAthena:
Increases flexibility with versioning

### DIFF
--- a/megatron/data/Makefile
+++ b/megatron/data/Makefile
@@ -1,3 +1,9 @@
+PYTHON3CONFIG := $(shell command -v python3-config 2> /dev/null)
+
+ifndef PYTHON3CONFIG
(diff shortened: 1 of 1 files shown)


megatron/data/Makefile
```diff
@@ -1,3 +1,9 @@
+PYTHON3CONFIG := $(shell command -v python3-config 2> /dev/null)
+
+ifndef PYTHON3CONFIG
```
Mistobaan, line 3: should it be `ifndef` or 
```bash
ifeq($(PYTHON3CONFIG),)
endif
```
I think the variable is always defined even if the shell command fails

StellaAthena: This made the code start working for me, so I don’t think the variable is always defined

StellaAthena:
Closing due to inactivity making this obsolete.

PR #508, opened 2022-02-09
Support for pytorch 1.10.2 and deepspeed 0.5.10

taylormcnally:
Some trivial changes to get gpt-neox working with deepspeed 0.5.10 and pytorch 1.10.2. Removed logits param and unsupported  mask method.

### DIFF
--- a/megatron/training.py
+++ b/megatron/training.py
@@ -205,7 +205,7 @@ def get_batch_pipe(data, neox_args):
 def forward_step(data_iterator, model, neox_args, timers, return_logits=False):
     """Forward step."""
     if neox_args.is_pipe_parallel:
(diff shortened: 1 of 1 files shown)


StellaAthena:
What have you tested with this, specifically? Multi-GPU? Multi-machine? Tensor parallelism? Pipeline parallelism?

taylormcnally:
I should have tested all of the configurations before making this, but so far I've tested single machine for small to medium sized models. I'm adding a GPU and testing multi-gpu and will get back the with the results

taylormcnally:
Tested with MP and PP for a multi GPU setup and it appears to be working. Though this is using a single machine. 

Screen Shot

PR #566, opened 2022-02-20
fix year

leogao2:


StellaAthena:
I noticed after merging the first PR that the model citation isn’t at the end with the other one. Is that deliberate? What was your thinking there?

StellaAthena (approved):

PR #572, opened 2022-02-23
fix bug in fused softmax

hyunwoongko:
I modified megatron's softmax code a while ago, and there were some mistakes in the first commit, so I made a few more commits to fix the bug. However, the code I accidentally committed was used in the NeoX side. This can cause serious problems.


ShivanshuPurohit (dismissed):

ShivanshuPurohit (approved):

hyunwoongko:
I can't understand why was CI failed...

PR #579, opened 2022-03-03
rocm 4.5.2 compliant

hyoo:
This is not meant to be merged directly. I just wanted to give an example of changes that you need to make to run on AMD GPUs (tested with rocm-4.5.2). Hope this is helpful.


StellaAthena (changes requested):
Declining so that this isn’t accidentally merged.

StellaAthena:
This has become the `AMD` branch, rather than merging it into `main`

PR #587, opened 2022-03-14
Fixed token_generation_end_index being too early

HughPH:
When using stop_tokens I found that output was cut off 1 token too early. Incrementing token_index_to_generate before setting token_generation_end_index solves this.


StellaAthena:
Looks like it's still failing :/

HughPH:
Not really a bug: a rolling 'chat room' context performed the role of few-shot examples and led my chat bot to deceive me into thinking a bug existed when it did not, and righted itself coincidentally on the next line after the introduction of the "fix". Fundamentally more thorough testing proved this to be faulty in premise.

Issue #589, opened 2022-03-16
[BUG] Can't install requirements with pip3 install -r requirements.txt command.


0x30c4:
**Describe the bug**
When I ran ```pip3 install -r requirements/requirements.txt``` it throws this error.
And same happens for the Docker image. 
It's probably related to some feature change.
https://github.blog/2021-09-01-improving-git-protocol-security-github/
```bash
root@t1v-n-7e10f855-w-0:/home/sanaf/neox/gpt-neox# pip3 install -r requirements/requirements.txt
Collecting deepspeed
  Cloning git://github.com/EleutherAI/DeeperSpeed.git (to revision eb7f5cff36678625d23db8a8fe78b4a93e5d2c75) to /tmp/pip-install-9w7hm3tg/deepspeed_5a2a7866c9094b508a1d1e935fd08969
  Running command git clone --filter=blob:none --quiet git://github.com/EleutherAI/DeeperSpeed.git /tmp/pip-install-9w7hm3tg/deepspeed_5a2a7866c9094b508a1d1e935fd08969
  fatal: remote error:
    The unauthenticated git protocol on port 9418 is no longer supported.
  Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
  error: subprocess-exited-with-error

  × git clone --filter=blob:none --quiet git://github.com/EleutherAI/DeeperSpeed.git /tmp/pip-install-9w7hm3tg/deepspeed_5a2a7866c9094b508a1d1e935fd08969 did not run successfully.
  │ exit code: 128
  ╰─> See above for output.

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× git clone --filter=blob:none --quiet git://github.com/EleutherAI/DeeperSpeed.git /tmp/pip-install-9w7hm3tg/deepspeed_5a2a7866c9094b508a1d1e935fd08969 did not run successfully.
│ exit code: 128
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.

```

EricHallahan:
Yes, it looks like the DeeperSpeed requirement doesn't use HTTPS like it should. I'll make a patch.

PR #590, opened 2022-03-16
[FIX] DeeperSpeed requirement doesn't use HTTPS.

0x30c4:
DeeperSpeed requirement doesn't use HTTPS but it should. So, I added a patch.

Here's the issue.
https://github.com/EleutherAI/gpt-neox/issues/589


StellaAthena:
I merged #591 which accomplishes the same thing. Appologies I didn't see your (chronologically prior) PR first Eric DM'd me about his.

0x30c4:
no it's ok!

PR #592, opened 2022-03-18
Add generating batch size

zhuzilin:
The origin implementation was generating samples one by one, which results in low gpu utility for small model. This commit adds a generating-batch-size config to enable batch generation.

Thank you for your time on reviewing this PR :)

### DIFF
--- a/configs/neox_arguments.md
+++ b/configs/neox_arguments.md
@@ -912,6 +912,14 @@ Text Generation arguments
 
 
 
+- **generating_batch_size**: int
+
+    Default = 1
+
+    The number of samples to generate at the same time
+
+
+
 ## NeoXArgsTokenizer
 
 Tokenizer Arguments
--- a/megatron/neox_arguments/neox_args.py
+++ b/megatron/neox_arguments/neox_args.py
@@ -974,3 +974,8 @@ class NeoXArgsTextgen(NeoXArgsTemplate):
     """
     Tasks to evaluate on using lm_eval_harness
     """
+
+    generating_batch_size: int = 1
+    """
+    The number of samples to generate at the same time.
+    """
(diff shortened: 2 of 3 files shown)


zhuzilin:
> I think a better solution is to drop the word train from the name of the existing argument and use the existing argument for generation. Having two seems needlessly confusing.

@StellaAthena I found that the `train_batch_size` arg relates to the `train_micro_batch_size_per_gpu`, which has something to do with gradient accumulation. And there is a `batch_size` arg used solely with  LM Evaluation Harness:
https://github.com/EleutherAI/gpt-neox/blob/1e346efe64c0d4ba04c6cfa6f8eb89679a6003d4/eval_tasks/eval_adapter.py#L70-L72

Maybe using the `batch_size` arg is better?

StellaAthena:
My suggestion is to rename `train_micro_batch_size_per_gpu` to `train_micro_batch_size_per_gpu` and allow it to fill both the train and the inference roles.

zhuzilin:
@StellaAthena 
Just a double check, you mean rename `train_batch_size` to `batch_size`, `train_micro_batch_size_per_gpu` to `micro_batch_size_per_gpu` and use `micro_batch_size_per_gpu` for inference/generation, right?

StellaAthena:
> @StellaAthena Just a double check, you mean rename `train_batch_size` to `batch_size`, `train_micro_batch_size_per_gpu` to `micro_batch_size_per_gpu` and use `micro_batch_size_per_gpu` for inference/generation, right?

No, I mean just the second rename. The training batch size and the eval batch size for a fixed microbatch size per gpu are typically different due to gradient accumulation.

zhuzilin:
@StellaAthena Updated.

zhuzilin:
@StellaAthena Hi, could you help me to fix the error raised in CI? Thank you!

PR #595, opened 2022-03-21
Add documentation generation to CI

slash-under:
This PR adds a new update documentation task, which runs the `gen_docs` script, then commits the changes as the `github-actions` user, to all new PRs, ensuring that any changes to NeoX arguments are updated in the documentation.

PR #596, opened 2022-03-22
Resolve whitespace issue that breaks the pull_request workflow

slash-under:
Resolves outstanding issues from #595, scope the `git add` call, install missing dependencies, and _temporarily_ remove two init files as they include unnecessary code.


StellaAthena:
Looks like this is still failing a test.

StellaAthena (approved):

PR #597, opened 2022-03-22
Load each stop_tokens token group into a Tensor

HughPH:
Evaluation of stop_tokens fails when is a jagged array, because a jagged array cannot be converted to a Tensor. Thus all token groups must be the same length. This precludes stop_tokens being passed as, for example, `[[187,187],[535]]` which are both representations of `\n\n`.

A _jagged array_ is an array of arrays, a vector of vectors or a list of lists where each member of the outer list is another list which may be of a different length to any or all of its siblings. Another example of a jagged array is `[[1], [2,3], [4,5,6]]`

This change makes any usage of stop_tokens with a 1-dimensional list incompatible. LMK if maintaining compatibility is desired, and I'll make the requisite changes.

### DIFF
--- a/megatron/text_generation_utils.py
+++ b/megatron/text_generation_utils.py
@@ -192,7 +192,7 @@ def stream_tokens(
     temperature: float = 0.0,
     top_k: int = 0,
     top_p: float = 0.0,
-    stop_tokens=None,
+    stop_tokens: List[List[int]]=None,
 ):
     """
     iterator producing text completions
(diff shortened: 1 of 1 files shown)


StellaAthena:
@HughPH The change is just to check if the list is one-dimensional and if so add an extra layer right? I think it's a good idea to add.

HughPH:
Yes, but the caveat would be un-typing the parameter again.

StellaAthena (approved):
Looks good to me.

PR #637, opened 2022-07-01
Add Support for Returning Logits for Generated Tokens

### LINKED ISSUE #588: Restore ability to get logits from generation?
It's often useful when generating samples to be able to get the logits / probabilities of the generated tokens (e.g. for ranking suggestions). It looks like this used to be available but was removed by 7aed13305c48b0048ec8b9efdba01e40fc69f123. It would be great if this functionality could be ported to the most recent version of the code.

I attempted to hack it back in, but must have messed something up because it made evaluation ~3.5x slower; I assume I accidentally introduced some GPU-CPU copies but unfortunately I don't really understand the Megatron/Deepspeed code well enough to see how to fix it. That branch is here: https://github.com/moyix/gpt-neox/tree/return_logits

Edit: fixed the link to my attempt at implementation, not sure how that happened

Kyle1668:
Closes #588

### DIFF
--- a/configs/neox_arguments.md
+++ b/configs/neox_arguments.md
@@ -871,6 +871,14 @@ Text Generation arguments
 
 
 
+- **return_logits**: bool
+
+    Default = False
+
+    Boolean for whether to return the logits for generated tokens
+
+
+
 - **maximum_tokens**: int
 
     Default = 64
@@ -914,7 +922,7 @@ Text Generation arguments
 
 - **eval_results_prefix**: str
 
-    Default = 
+    Default =
 
     prefix to which to save evaluation results - final fp will be {eval_results_prefix}_eval_results_yy-mm-dd-HH-MM.json
 
@@ -1502,7 +1510,6 @@ Args for deepspeed config
 
     Default = None
 
-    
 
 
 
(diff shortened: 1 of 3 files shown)


Kyle1668:
The original issue mentioned potentially returning the log-likelihood of the tokens through some integration in the `generate.py` interface. I decided against implementing that support in this PR and instead returned the raw logits. My reasoning is that end users can then operate on these logits as needed.

Kyle1668:
Outputs from ` ./deepy.py generate.py ./configs/20B.yml ./configs/text_generation.yml` on my branch

With `return_logits: true` 

image

without 

```json
{
  "context": "",
  "text": "Entry DescriptionWild Agnes has released this extraordinary new",
  "length": 10,
  "finished": false,
  "message": null,
  "duration_seconds": 6.626634120941162
}
```

on main

```json
{
  "context": "",
  "text": "Entry DescriptionWild Agnes has released this extraordinary new",
  "length": 10,
  "finished": false,
  "message": null,
  "duration_seconds": 6.523743629455566
}
```

Kyle1668:
I took a stab at writing some unit tests for this change but ran into an issue with `mpu.get_model_parallel_src_rank()` throwing an exception saying that I needed to configure `init_process_group`.

StellaAthena (approved):
This is great! Apologies that this PR took so long to review and merge, but we hugely appreciate this work.

PR #652, opened 2022-07-29
Update AliBi matrix caching for dynamic sequence lengths

VHellendoorn:
When using a model trained with AliBi positional embedding for inference, the cached matrix gets invalidated and recomputed after every generated token, which is very expensive. This PR offers an alternative where the matrix is grown exponentially when longer sequences are encountered, while using clipping to ensure it works for shorter sequences. This will have no effect at training time, where it will be initialized to, and remain unchanged at, the target sequence length as before.

It may also be worth setting a minimum target length, e.g. at 64 tokens, possibly making this depend on the max length specified in the text generation config. In any case, the below has been tested with an AliBi-trained model and allows fast generation after the initial few resizing steps.

### DIFF
--- a/megatron/model/positional_embeddings.py
+++ b/megatron/model/positional_embeddings.py
@@ -125,9 +125,10 @@ class AliBi(torch.nn.Module):
 
         # Initialize the AliBi matrix to match the first provided key length; grow it exponentially
         # afterwards if longer inputs are provided. This is important for inference, where we will
-        # encounter progressively longer samples, while incurring no additional cost when input
-        # sequences are all equally long, as is typical at training time.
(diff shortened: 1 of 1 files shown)


StellaAthena:
This sounds like a really valuable correction! Thanks for the contribution. I have a lot on my plate and am not totally sure when I’ll have the bandwidth to review this PR but it’s a high priority :)

VHellendoorn:
Sounds good! No problem; I am using it in my fork for now. I also just realized the inital PR version had a wrong condition that I'd fixed locally (hence the new commit), so all the better that it got some time to sit :)

StellaAthena (approved):

PR #665, opened 2022-09-04
Updating default configs to be less bad

StellaAthena:


haileyschoelkopf:
@StellaAthena Since we've trained Pythia models at ~all but the 175B sizes, should we merge this? I think external users frequently reference the current main branch default configs when using this library for their own training.

StellaAthena:
@haileyschoelkopf good call. Honestly, I had forgotten that we didn’t. However it seems like there’s merge conflicts I can’t do on my phone, so it’ll have to wait until the morning.

Quentin-Anthony:
@haileyschoelkopf -- Can you also add the Pythia model configs (800M, 49M, etc) to this branch before we merge?

haileyschoelkopf:
of course, will do later today!

haileyschoelkopf:
I've added these configs! (All Pythia configs are also in the Pythia repo.) 

Should be good to merge!

Quentin-Anthony (approved):

PR #667, opened 2022-09-05
Resolve `hostfile` CLI Param Overwriting Config Value

### LINKED ISSUE #666: The `hostfile` NeoX Argument Is No Longer Working
**Describe the bug**

The `hostile` NeoX configuration argument is no longer taking effect. Regardless of the value set, the hostile path is always `/job/hostfile`.

image

The hostile updates if you use the `--hostfile` parameter. 

image

**To Reproduce**
1. On the latest main build, run `pytest tests -m cpu` and observe that the `test_neoxargs_consume_deepy_args_without_yml_suffix` unit test is failing.
2. (Alt) Try to set a custom value for the `hostfile` config and observe that the program nonetheless uses the `/job/hostfile` path.

**Expected behavior**

When the `hostfile` argument is set, the value used in the program (even if it is a dummy path) matches the config value.

image

The `test_neoxargs_consume_deepy_args_without_yml_suffix` unit test passes.

image

**Proposed solution**

I suspect that this bug was introduced in 5e0d614744983cc87cb33eb127c60f13562399bc . When I comment out the CLI parameter, the test passes. 

image

The default value for this parameter is likely overwriting any values read from the config. 

**Environment (please complete the following information):**
 - GPUs: 8x NVIDIA A40
- Configs: Default
- Environment: Ubuntu in EleutherAI Cluster

Kyle1668:
Closes #666

### DIFF
--- a/configs/small.yml
+++ b/configs/small.yml
@@ -81,4 +81,7 @@
    "steps_per_print": 10,
    "keep-last-n-checkpoints": 4,
    "wall_clock_breakdown": true,
+
+  #  networking
+  "hostfile": "/mock_path"
 }
--- a/megatron/neox_arguments/arguments.py
+++ b/megatron/neox_arguments/arguments.py
@@ -288,7 +288,6 @@ class NeoXArgs(*BASE_CLASSES):
             "-H",
             "--hostfile",
             type=str,
-            default=DLTS_HOSTFILE,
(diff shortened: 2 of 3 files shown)


Kyle1668:
image

Passing Tests

StellaAthena (approved):

PR #676, opened 2022-09-18
Multitask finetuning

lintangsutawika:
I've re-implemented the original Super Glue prompts from the T5 Paper. Currently reconciling MTF training process with improved-t5-2.0 (Since the finetuning on SuperGLUE for T5 is exactly the same as Multitask Finetuning)

cc: @haileyschoelkopf  @AranKomat


megatron/data/data_utils.py
```diff
@@ -74,49 +74,27 @@ def build_the_dataset(
...
-            data_impl,
-            skip_warmup=False,
-            tokenizer=neox_args.tokenizer,
-        ) 
-    elif neox_args.training_objective == "mlm":
-
-        dataset = MLMDataset(
-            *dataset_args,
+
+    if neox_args.model_arch == "gpt2":
```
StellaAthena, line 78: Is there a particular reason to use GPT2 here over GPT3? I know that they're the same but I wouldn't necessarily assume all users know that.

lintangsutawika: No strong reason, only done for the sake of naming consistency since it will call the model class object from gpt2_model.py

StellaAthena: lol I thought I had bumped those up to GPT-3 as well.. for now, this is the right choice (if only for consistency)

PR #680, opened 2022-09-21
SLURM Multi-Node Support

dashstander:
Pairs with [this](https://github.com/EleutherAI/DeeperSpeed/pull/40) DeeperSpeed PR.

The big thing to pay attention to is that when using SLURM and `srun` python's argument parser becomes a lot more picky. The most expedient way to solve this is, when using SLURM, to modify your config files to conform to strict JSON. The regular configs, despite their `yml` extensions, are already pretty close to JSON except for comments and a few capital-T `True`s. There is also an [informative error](https://github.com/EleutherAI/DeeperSpeed/blob/23797d3b704a070d6a9ce5d680e4189e59e0e101/deepspeed/launcher/multinode_runner.py#L143) if someone runs afoul of this limitation.

I think that this is an acceptable solution, though I understand if people would rather not fracture configs based on cluster type in this way. If people have serious reservations I'm open to discussing other options.

There are also small functional changes regarding correctly finding SLURM environment variables.


StellaAthena:
@Quentin-Anthony or @ShivanshuPurohit can you review this and give it a thumbs up?

configs/slurm_local.yml
```diff
@@ -0,0 +1,13 @@
+{
+  "data-path": "data/enron/enron_text_document",
+  "vocab-file": "data/gpt2-vocab.json",
+  "merge-file": "data/gpt2-merges.txt",
+  "save": "checkpoints",
```
Quentin-Anthony, line 5: Add a comment explaining why this is different from `configs/local_setup.yml` (i.e. why you capitalized bools)

Quentin-Anthony: Oops that's not possible. Comments aren't parsed correctly by srun. We should document this, but elsewhere.

megatron/neox_arguments/arguments.py
```diff
@@ -544,6 +544,12 @@ def configure_distributed_args(self):
             from deepspeed.utils.distributed import mpi_discovery
 
             mpi_discovery()
+        
+        # TODO: Should there just be an argument / attribute like self.deepspeed_mpi for this block?
```
Quentin-Anthony, line 548: Add `deepspeed_slurm` to guard this block. Replace the `os.getenv` with it.

dashstander:
Made @Quentin-Anthony 's changes. The pre-commit hooks pass on my local...
Screen Shot 2022-09-21 at 2 59 20 PM

Quentin-Anthony (approved):

PR #697, opened 2022-09-28
Bigscience lm-eval-harness fork support

haileyschoelkopf:
This PR updates the interface to `lm-eval-harness` used to support the Promptsource-integrated Bigscience fork ([see fork here](https://github.com/bigscience-workshop/lm-evaluation-harness) for more info) and changes logging to support it as well. At the moment template selection and evaluation on multiple tasks at once is not supported, it will be.

Changes here are relatively minor, but **should not be merged because it is not backwards compatible with the upstream eval-harness.**


haileyschoelkopf:
Addresses issue #697 .

PR #701, opened 2022-10-06
NeoX to HF Model Conversion Script

haileyschoelkopf:
[WIP still] This is a script to convert models trained in NeoX to Huggingface format for uploading to the hub. This should work as is, but draft for now.

I'll add as well the actual HF `push_to_hub` commands to this script to try to semi-automate pushing models to the hub, ideally.

Note that this won't work for any feature unsupported by HF's NeoX implementation at the moment (AliBi, for one). We'd have to write and get HF to merge those extra architecture changes

@StellaAthena @LouisCastricato


haileyschoelkopf:
@Quentin-Anthony This should be ready for review! 

Note that it requires bumping the HF package versions to access the GPT-NeoX arch and the newer `use_parallel_residual` config option.

StellaAthena:
> @Quentin-Anthony This should be ready for review!
> 
> Note that it requires bumping the HF package versions to access the GPT-NeoX arch and the newer `use_parallel_residual` config option.

Sid was adamant about using strictly single-version requirements specifications but I don’t see any reason this should be problematic (also, I don’t agree with Sid about this)

tools/convert_to_hf.py
```diff
@@ -0,0 +1,271 @@
...
+
+def convert(input_checkpoint_path, loaded_config, output_checkpoint_path):
+    """ convert a NeoX checkpoint to a HF model format. 
+    should perform model-parallel merging correctly
+    but only supports features allowed by HF GPT-NeoX implementation (e.g. rotary embeddings)
+    """
+
+    hf_config = GPTNeoXConfig()
+
+    # TODO: how to deal with vocab size? need to pass tokenizer and calc on the fly.
```
Quentin-Anthony, line 108: Has this TODO been resolved? If so, let's remove this. If not, I'm not sure what "need to pass tokenizer and calc on the fly" means

haileyschoelkopf: Yes sorry, removed it!

This refers to calculating the vocab size with padding to a multiple of 128, and it’s done in this script now.

Quentin-Anthony:
> > @Quentin-Anthony This should be ready for review!
> > Note that it requires bumping the HF package versions to access the GPT-NeoX arch and the newer `use_parallel_residual` config option.
> 
> Sid was adamant about using strictly single-version requirements specifications but I don’t see any reason this should be problematic (also, I don’t agree with Sid about this)

I agree. This looks good to me. Thanks @haileyschoelkopf!

Quentin-Anthony (approved):

PR #704, opened 2022-10-19
Mup Support

nsarka:
This PR implements support for mup: https://github.com/microsoft/mup

I'm able to train with this patch. Studies on its effectiveness still need to be done. For that, I still need to add args for generating the coord check plots they describe in the readme.

### DIFF
--- a/configs/small.yml
+++ b/configs/small.yml
@@ -7,7 +7,6 @@
    # across the node boundaries )
    "pipe-parallel-size": 1,
    "model-parallel-size": 1,
-   "is-pipe-parallel": true,
 
    # mup
    "use-mup": true,
--- a/configs/small.yml
+++ b/configs/small.yml
@@ -10,6 +10,8 @@
 
    # mup
    "use-mup": true,
+   "save-base-shapes": true, # this only needs to be enabled once in order to generate the base-shapes-file on each rank
+   "base-shapes-file": "base-shapes", # load base shapes from this file
 
    # model settings
    "num-layers": 12,
--- a/configs/small.yml
+++ b/configs/small.yml
@@ -7,6 +7,7 @@
    # across the node boundaries )
    "pipe-parallel-size": 1,
    "model-parallel-size": 1,
+   "is-pipe-parallel": true,
 
    # mup
    "use-mup": true,
--- a/configs/small.yml
+++ b/configs/small.yml
@@ -1,9 +1,15 @@
 # GPT-2 pretraining setup
 {
+  "launcher": "openmpi",
+  "deepspeed_mpi": true,
+
    # parallelism settings ( you will want to change these based on your cluster setup, ideally scheduling pipeline stages
    # across the node boundaries )
    "pipe-parallel-size": 1,
    "model-parallel-size": 1,
+
+   # mup
+   "use-mup": true,
 
    # model settings
    "num-layers": 12,
@@ -16,8 +22,8 @@
    "no-weight-tying": true,
 
    # these should provide some speedup but takes a while to build, set to true if desired
-   "scaled-upper-triang-masked-softmax-fusion": false,
-   "bias-gelu-fusion": false,
+   "scaled-upper-triang-masked-softmax-fusion": true,
+   "bias-gelu-fusion": true,
 
 
    # optimizer settings
@@ -30,7 +36,7 @@
      }
    },
    "zero_optimization": {
-    "stage": 0,
+    "stage": 1,
     "allgather_partitions": True,
     "allgather_bucket_size": 500000000,
     "overlap_comm": True,
@@ -41,7 +47,7 @@
   },
 
    # batch / data settings
-   "train_micro_batch_size_per_gpu": 4,
+   "train_micro_batch_size_per_gpu": 32,
    "data-impl": "mmap",
    "split": "949,50,1",
 
@@ -77,7 +83,7 @@
    "eval-iters": 10,
 
    # logging
-   "log-interval": 100,
+   "log-interval": 10,
    "steps_per_print": 10,
    "keep-last-n-checkpoints": 4,
    "wall_clock_breakdown": true,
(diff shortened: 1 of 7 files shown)


megatron/training.py
```diff
@@ -297,6 +350,9 @@ def get_optimizer(model, neox_args):
         _param_groups.append(param_group)
     param_groups = _param_groups
 
+    # If we're using mup, then the optimizer must be adam or sgd
+    assert not neox_args.use_mup or (neox_args.optimizer_type.lower() == "adam" or neox_args.optimizer_type.lower() == "sgd")
```
Quentin-Anthony, line 354: Let's print a message if this assertion fails so that they don't have to read this comment. 

assert not neox_args.use_mup or (neox_args.optimizer_type.lower() == "adam" or neox_args.optimizer_type.lower() == "sgd"), f'If use_mup == True, you must specify either the adam or sgd optimizers. You passed: {neox_args.optimizer_type.lower()}'

megatron/training.py
```diff
@@ -234,6 +277,8 @@ def get_model(neox_args, use_cache=False):
     print_rank_0("building GPT2 model ...")
 
     # Build model on cpu.
+    old_use_mup = neox_args.use_mup
```
Quentin-Anthony, line 280: I assume the point of old_use_mup logic is to ensure we aren't using mup optimizers for the model? Add a comment to that effect if so. Explain why we use this logic if not.

configs/small.yml
```diff
@@ -1,10 +1,18 @@
 # GPT-2 pretraining setup
```
Quentin-Anthony, line 1: Unstage the config changes before merging

Quentin-Anthony:
I left a set of comments. Let's resolve those, test changes with PP, get some coord check plots, and test a few other mup settings + neox config combinations before merging.

megatron/training.py
```diff
@@ -57,6 +57,49 @@
...
+
+    if not neox_args.is_pipe_parallel:
+        base_model = base_model.to_sequential()
+
+    base_shapes = mup.get_shapes(base_model)
+
+    del base_model
+
+    neox_args.hidden_size = neox_args.hidden_size * 2
+    neox_args.ffn_hidden_size = 4 * neox_args.hidden_size
```
Quentin-Anthony, line 82: `ffn_hidden_size` and `kv_channels` don't exist in neox, so these two lines aren't doing anything.

megatron/training.py
nsarka: I added a comment explaining the original reason I added the old_use_mup logic. After looking at the docs again, I need to add another function that will reinitialize the model's weights using mup's functions (since setting use_mup to false here will make the model initialize with torch.nn's functions). Then, call that function after the weights were reinitialized.

megatron/training.py
```diff
@@ -57,6 +57,49 @@
...
+                    use_cache=use_cache)
+
+    if not neox_args.is_pipe_parallel:
+        base_model = base_model.to_sequential()
+
+    base_shapes = mup.get_shapes(base_model)
+
+    del base_model
+
+    neox_args.hidden_size = neox_args.hidden_size * 2
```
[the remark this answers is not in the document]
nsarka: I think if megatron/neox_arguments/neox_args.py supports taking in a list and we identified which hparams are good candidates for scaling, it wouldn't be too much effort to edit the code for that.

Also, what you set each hparam to in the delta model might not matter. From the docs it just says you need to change whatever you want to scale. It might be worth double checking that with the authors though.

megatron/model/init_functions.py
```diff
@@ -2,28 +2,40 @@
 
 import torch
 
+try:
```
Quentin-Anthony, line 5: This import will still fail. Mup cannot be a top-level import (we only want an exception if (mup is not installed) and (we try to use mup)

StellaAthena: What precisely is wrong with making MuP a requirement? Like, under what conditions is expecting someone to have access to it unreasonable? Or does it have negative externalities I am unaware of?

Quentin-Anthony: My argument is to reduce requirements bloat. I don't expect many users to require mup, so I don't want it to be required. I consider mup to be similar to our bnb support: "A specialized optimization that's mostly for advanced internal users"

megatron/training.py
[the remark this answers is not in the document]
Quentin-Anthony: Ok. Let's revisit this once the PR is more mature.

megatron/training.py
Quentin-Anthony: We discussed this offline. @nsarka is going to add a function to reinit all model weights if mup is enabled after running `set_base_shapes()`. The expected implementation method will be a for-loop over layers modeled after `to_sequential()`

nsarka:
I updated the PR to include a neox argument for generating coord check plots. Here's the output:

Screen Shot 2022-11-06 at 7 10 56 PM

The bottom row has the charts for standard parameterization (SP, meaning no mup) showing that with each increase of width, the mean output from each layer is increasing. The top row shows mup makes it constant which means we can transfer.

The only thing that seems odd to me is that the SP chart seems to not change with each t (one call to train_step in neox). In the example plots from mup, the curves start off flat and with each t the slopes get steeper. In these charts they start out steep and stay the same.

For now I'm assuming it's not a problem and will start some transfer experiments.

megatron/training.py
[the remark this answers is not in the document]
StellaAthena: @nsarka @Quentin-Anthony now that the PR is mature, we should revisit this conversation.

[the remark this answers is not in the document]
Quentin-Anthony: @StellaAthena -- We discussed this offline and added a new arg `mup_width_scale` to scale the hidden size (default = 2). If we find scaling other hparams provides benefits we can add those in the future.

PR #709, opened 2022-10-29
Allow building of fused kernels without GPU

salanki:
The pytorch based CUDA check in the setup script of the Megatron fused kernels would return False if a GPU is not available. A GPU is not required to build the fused kernels. This check prevents properly building NeoX image in CI.

### DIFF
--- a/megatron/fused_kernels/setup.py
+++ b/megatron/fused_kernels/setup.py
@@ -1,7 +1,6 @@
 from setuptools import setup, find_packages
 from torch.utils import cpp_extension
 from torch.utils.cpp_extension import BuildExtension, CUDAExtension
-from torch.cuda import is_available as torch_cuda_available
 from pathlib import Path
 import subprocess
 
--- a/megatron/fused_kernels/setup.py
+++ b/megatron/fused_kernels/setup.py
@@ -64,8 +64,6 @@ setup(
             ],
             extra_compile_args=cuda_ext_args,
         ),
-    ]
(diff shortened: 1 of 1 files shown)


Quentin-Anthony:
@salanki -- I agree with you. 

**Some discussion for those interested (and future maintainers):**

Building a PyTorch CUDAExtension requires at least one of the following:

1. Define the `TORCH_CUDA_ARCH_LIST` env var (https://github.com/pytorch/pytorch/blob/v1.13.0/torch/utils/cpp_extension.py#L1759)
2. Have at least one GPU present when building (https://github.com/pytorch/pytorch/blob/v1.13.0/torch/utils/cpp_extension.py#L1766-L1788)
3. Pass extra compile flags containing the target arch (https://github.com/pytorch/pytorch/blob/v1.13.0/torch/utils/cpp_extension.py#L1731-L1736)

Since gpt-neox builds with option 3: 

https://github.com/EleutherAI/gpt-neox/blob/1cfdebf42d4fdd1548564d2b1785d0411a31908d/megatron/fused_kernels/setup.py#L25-L38

We can forego the extra `torch.cuda.is_available()` check. Note that this assumes that the CUDA used to build fused kernels matches the major version to be used on GPU.

For my own remembrance, PyTorch's cc logic is discussed in https://github.com/pytorch/pytorch/blob/v1.13.0/torch/utils/cpp_extension.py#L975-L1004

Quentin-Anthony (approved):

PR #719, opened 2022-11-23
Allow wildcard-style Eval Harness task selection

haileyschoelkopf:
The lm-eval-harness uses Linux wildcards to allow for the selection of multiple tasks sharing a prefix in their name at once, via passing `--tasks blimp_* ` to select all BLiMP tasks at once rather than manually.

This PR adds support for this pattern to the Eval-harness adapter, so that one can pass ` --eval_tasks blimp_* ` or `--eval_tasks hendrycks_* ` to run on all BLiMP or MMLU tasks without passing all task names individually, as a minor QoL fix.

### DIFF
--- a/configs/neox_arguments.md
+++ b/configs/neox_arguments.md
@@ -111,7 +111,7 @@ Logging Arguments
 
 - **git_hash**: str
 
-    Default = 2b001f8
+    Default = 32deba2
 
     current git hash of repository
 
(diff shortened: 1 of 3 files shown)


Quentin-Anthony (approved):

Quentin-Anthony:
Thanks for this, @haileyschoelkopf!

Issue #721, opened 2022-11-25
Broken "learned" embeddings


Quentin-Anthony:
This PR (https://github.com/EleutherAI/gpt-neox/commit/7c25eed8a8d5a6a87560370f2a6187f138eca0af#diff-5ae2f919cc2b3b2cf7bc8193b6e1c1cfd2e9c15d75edcb44b213ccc80b772110R122) breaks "learned" embeddings with:

```
  File "/fsx/quentin/gpt-neox/megatron/model/utils.py", line 187, in forward
    x = exec_range_func(start_idx, end_idx)(*x)
  File "/fsx/quentin/gpt-neox/megatron/model/utils.py", line 160, in exec_func
    inputs = layer(inputs)
  File "/fsx/quentin/miniconda3/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/fsx/quentin/gpt-neox/megatron/model/word_embeddings.py", line 162, in forward
    embeddings = super().forward(input_ids, position_ids)
  File "/fsx/quentin/gpt-neox/megatron/model/word_embeddings.py", line 124, in forward
    position_ids = position_ids + self.layer_past + 1
RuntimeError: The size of tensor a (2048) must match the size of tensor b (4) at non-singleton dimension 1
```

Can be replicated with 125M.yml local_setup.yml and enron on a single node, but I assume most configs will set it off. Reverting https://github.com/EleutherAI/gpt-neox/commit/7c25eed8a8d5a6a87560370f2a6187f138eca0af#diff-5ae2f919cc2b3b2cf7bc8193b6e1c1cfd2e9c15d75edcb44b213ccc80b772110R121-R129 resolves the issue.

StellaAthena:
@Quentin-Anthony did you test if the OPT-style positional embeddings still work with your fix? Personally, I’m not particularly attached to allowing people to use them, but if they don’t we should probably just remove the config entirely.

Tagging @zphang for visibility

Quentin-Anthony:
> @Quentin-Anthony did you test if the OPT-style positional embeddings still work with your fix? Personally, I’m not particularly attached to allowing people to use them, but if they don’t we should probably just remove the config entirely.
> 
> Tagging @zphang for visibility

[...]

PR #722, opened 2022-11-27
Fix learned and sinusoidal embeddings after OPT regression

Quentin-Anthony:
Fixes https://github.com/EleutherAI/gpt-neox/issues/721

@zphang -- Can you test this with OPT?

@StellaAthena -- FYI


StellaAthena:
@zphang did you ever have a chance to review this?

dashstander:
I've tested this and it doesn't break anything in the normal workflow @Quentin-Anthony

PR #728, opened 2022-12-02
Raise error in HF conversion script if using single tied layernorm with GPT-J residual

haileyschoelkopf:
This PR makes the conversion to HF format raise an error if a model with a single layernorm and GPT-J residual setup is used, since it's not currently supported by HF.

cc @Quentin-Anthony @StellaAthena


StellaAthena (approved):
Looks good to me!

haileyschoelkopf:
Merged!

PR #739, opened 2022-12-09
Implement DeepSpeed Main autotuning for NeoX

dashstander:
Implements [autotuning](https://github.com/microsoft/DeepSpeed/tree/master/deepspeed/autotuning) feature from DeepSpeed, which programmatically explores different combinations of microbatch size, gradient accumulation, and ZeRO sharding options given your model and compute set up.


dashstander:
The slowdown is due to a bug in the code [DeepSpeed autotuning uses to report FLOPS](https://github.com/EleutherAI/DeeperSpeed/blob/a30e28aa4c3f003c1b7d570231a45baee76c6fbe/deepspeed/runtime/engine.py#L2277).  The latency is in milliseconds and are not converted to seconds when recording the FLOPS. We will be making a PR to DeepSpeed to fix this issue upstream. Now this PR is very close to complete, I'll do a little clean up this evening and then it will be camera-ready.

dashstander:
Ok, should be ready for review @Quentin-Anthony

Quentin-Anthony (approved):

PR #754, opened 2022-12-22
Enable multiline prompts by setting a custom prompt end

kabachuha:
Available both for input-file and interactive modes, it rerolls the input request in the latter until it meets the stop sequence.

Especially useful for continuing indent-aware code and lists generation.

I tested it for both cases and it works as intended

### DIFF
--- a/megatron/text_generation_utils.py
+++ b/megatron/text_generation_utils.py
@@ -746,7 +746,7 @@ def generate_samples_interactive(
             raw_text = ""
             while True:
                 current_input = input("Context prompt >>> ")
-                if prompt_end == "\n":
+                if prompt_end == "\n": # we need to handle '\n' case as 'input' strips it and leads to lines being squashed
                     raw_text += current_input
                     break
                 if prompt_end in current_input:
@@ -754,7 +754,7 @@ def generate_samples_interactive(
                     break
                 raw_text += (
                     current_input + "\n"
-                )  # read newline since we stripped it on input
+                )  # re-add newline since we stripped it on input
             context_tokens = neox_args.tokenizer.tokenize(raw_text)
             if len(context_tokens) == 0:
                 context_tokens = [neox_args.tokenizer.eod]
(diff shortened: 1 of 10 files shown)


megatron/text_generation_utils.py
```diff
@@ -738,7 +743,18 @@ def generate_samples_interactive(
 
         if torch.distributed.is_initialized() and torch.distributed.get_rank() == 0:
             os.system("clear")
-            raw_text = input("Context prompt >>> ")
+            raw_text = ""
```
StellaAthena, line 746: Can you elaborate on why this code looks like like it treats `\n` as a special case?

megatron/text_generation_utils.py
```diff
@@ -738,7 +743,18 @@ def generate_samples_interactive(
...
+            while True:
+                current_input = input("Context prompt >>> ")
+                if prompt_end == "\n":
```
StellaAthena, line 749: Can you elaborate on why this part of the code treats `\n` as a special case?

kabachuha: python's `input` function strips the newline character, so while it reads multiple lines it combines them into a single line input unless it is returned back
I got that behavior when testing, so it's not a speculation

StellaAthena: Ah I see. I had forgotten about that and assumed you were handling the default case separately for some reason.

Could you add a note about this for future maintainers? Otherwise it looks good and if it tests well I’ll merge it.

kabachuha: Added them now in [437bbf5](https://github.com/EleutherAI/gpt-neox/pull/754/commits/437bbf5d42b4df14a43eed4a9db97aebdc67d82d)

Quentin-Anthony:
Thanks for this feature @kabachuha!

I went ahead and cleaned up the formatting/spelling of this PR and some other recent commits. The checks here are still failing, but pass for me, so there's likely some issues on our runner. I'll merge anyway and look into that later.

Quentin-Anthony (approved):

PR #756, opened 2022-12-25
set gpt_j_tied to true so that 20B inference works again

chuanli11:
`gpt_j_tied=false` will cause 20B model generate gibberish results. This is reported [here](https://github.com/EleutherAI/gpt-neox/issues/745) and [here](https://github.com/EleutherAI/gpt-neox/issues/755).

The source of the problem is this [commit](https://github.com/EleutherAI/gpt-neox/commit/0accac618d70bd62de3965c037a7b2a36d07deb1).

To reproduce, run `./deepy.py generate.py ./configs/20B.yml`

"gpt_j_tied": false
```
{"context": "", "text": " de de general de general de2a1a\u00ada\u00ada\u00ada\u00ada\u00ada\u00ada\u00ado\u00ado\u00ado\u00ado\u00ado\u00ado\u00ado\u00ado\u00ado\u00ado3\u00ado\u00ado3\u00ado3\u00ado3\u00ado\u00ado\u00ado\u00ado\u00ado", "length": 64, "finished": false, "message": null, "duration_seconds": 4.648434400558472}
```
"gpt_j_tied": true

```
{"context": "", "text": "Q:\n\nHow to get the value of a variable in a function in a different file?\n\nI have a file called \"main.py\" and a file called \"functions.py\".\nIn \"main.py\" I have:\nfrom functions import *\n\ndef main():\n    print", "length": 64, "finished": false, "message": null, "duration_seconds": 4.764044523239136}
```

PR #761, opened 2022-12-27
Fix behavior of `gpt_j_tied` synced LayerNorms

haileyschoelkopf:
This PR makes it so that when `neox_args.gpt_j_tied = True`, the layernorms are tied to each other. When `neox_args.gpt_j_tied = False`, there are now 2 separate untied layernorms as intended.


StellaAthena (approved):
Thanks for catching this!

This PR corrects https://github.com/EleutherAI/gpt-neox/pull/756

PR #787, opened 2023-02-14
Add DeepSpeed bf16 configuration

dashstander:
Let's try this again: DeepSpeed uses a special `bf16` configuration to enable using bf16 precision during. This adds that configuration to NeoX and main DeepSpeed does the rest of the work.

### DIFF
--- a/megatron/neox_arguments/deepspeed_args.py
+++ b/megatron/neox_arguments/deepspeed_args.py
@@ -89,6 +89,11 @@ class NeoXArgsDeepspeedConfig(NeoXArgsTemplate):
     Configuration for using mixed precision/FP16 training that leverages NVIDIA’s Apex package.
     """
 
+    bf16: dict = None
+    """
+    Configuration for using bfloat16 precision during training.
+    """
+
     amp: dict = None
     """
     Dictionary as described in Deepspeed documentation: https://www.deepspeed.ai/docs/config-json/#automatic-mixed-precision-amp-training-options


configs/neox_arguments.md
```diff
@@ -1680,6 +1680,14 @@ Args for deepspeed config
 
 
 
+- **bf16**: dict
```
StellaAthena, line 1683: We already support the `precision` config option. Instead of making a new config for bf16, let's make `bf16` an allowable value for `precision` to take on.

StellaAthena (changes requested):

StellaAthena:
I think at it’s core my question is whether this is actually more intuitive than:
```
precision = 'bfloat16'
precision_args = {whatever you need goes here}
```

dashstander:
As things currently work we can't _just_ have a `precision_args` argument that gets turned into  a `bf16` or `fp16` dict for DeepSpeed because every argument that gets passed to DeepSpeed needs to exist in the `megatron/neox_args/deepspeed_args.py` file.

The changes I've made (and didn't push  all of last night, whoops) make it so that a user only has to set the `precision` argument and the bare-bones of everything else gets set. It also raises an error if those dictionaries are set but the `precision` argument disagrees.

I'm open to other approaches though.

dashstander:
I still need to test this to confirm there aren't any stupid bugs, but otherwise @StellaAthena and @Quentin-Anthony I'd appreciate if you took a look at this when you have a moment.

If people still don't like this--and honestly I don't _love_ it--then I think we should open another issue where we figure out a way for people pass stuff straight into DeepSpeed without getting checked by NeoX or needing to have it be checked into our code. I know we've talked about it before. In that scenario we'd be able to just handle the default cases when people set `precision` and for advanced users who want to mess with the `fp16` args dict they can do so.

eval_tasks/eval_adapter.py
```diff
@@ -30,6 +30,7 @@ def _download_file(*args, **kwargs):
 import sys
 import dataclasses
 from functools import partial
+from pathlib import Path
```
Quentin-Anthony, line 33: What's this change for?

configs/neox_arguments.md
```diff
@@ -1766,16 +1774,6 @@ Args for deepspeed config
 
 
 
-- **bf16**: dict
-
-    Default = None
-
-    Configuration for using bfloat16 floating-point format as an alternative to FP16. BFLOAT16 requires hardware support (e.g., NVIDIA A100).
```
Quentin-Anthony, line 1773: Why'd we remove these details?

dashstander: They're tied to the specific `bf16` config that you wanted removed. I can move them to the section on the `precision` config?

Quentin-Anthony (approved):

PR #795, opened 2023-02-22
Remove need/use of password.txt

slycordinator:
Removes redundant generation/use/deletion of password.txt by placing the password as a variable and running echo with the variable directly.

Previously, the text 'password' was echoed into password.txt, then read from it again, before the file would finally get deleted.


Quentin-Anthony (approved):

Quentin-Anthony:
Thanks for this!

PR #799, opened 2023-02-27
Update corpora urls

HuangLK:
`mystic.the-eye.eu` is down, should replace it with `the-eye.eu`.


StellaAthena (approved):

StellaAthena:
I really thought we had previously fixed these links. Thanks!

PR #802, opened 2023-02-27
added a simple script for multi-node data preparation.

MLRichter:
I added a simple bash script for multi-node data preparation.

**Why is it needed**
Large amounts of data may require more processing than a single compute-node can provide within a reasonable amount of time.
Therefore, multi-node support would enable an easy and straightforward way to preprocess the data.

**How does it work**
A list of files containing the data is provided. These files a split up evenly across the different ranks of the multi-node setup. Each rank thus processes a chunk of the data.

**What needed to change?**
Nothing in the python code itself has changed. However, there is a bash-script now that can be called in your distributed setup of choice instead of `tools/preprocess_data.py`

### DIFF
--- a/tools/multinode_prepare_data.sh
+++ b/tools/multinode_prepare_data.sh
@@ -43,7 +43,7 @@ done
 
 # Add the rank to the --output-prefix argument if it is set
 if [[ -n "$prefix_arg" ]]; then
-    py_args="$py_args $prefix_arg$rank)"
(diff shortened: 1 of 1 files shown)


StellaAthena:
This looks correct, but I'll need to have someone on our team test it a bit before we can merge. That said, thanks for the contribution!

Quentin-Anthony:
Tested and working fine.

Quentin-Anthony (approved):

PR #803, opened 2023-02-28
Remove password.txt in Dockerfile

xu-song:
password.txt is a redundant file


Quentin-Anthony:
Resolved in https://github.com/EleutherAI/gpt-neox/pull/795

PR #806, opened 2023-03-03
Only import tiktoken if we intend to use it

Quentin-Anthony:


StellaAthena (approved):
Seems reasonable enough

PR #819, opened 2023-03-09
Extra DeepSpeed Argument Capability

curt-tigges:
Added functionality to take in all additional DeepSpeed arguments through the YAML configuration files, using the deepspeed_extra_args option. Will validate to ensure that arguments are not duplicated elsewhere, and then add arguments to those sent to DeepSpeed.


Quentin-Anthony (approved):
Looks good! Tested and approved.

PR #824, opened 2023-03-10
Update fused kernel error recommendation, and make error logs more verbose

Quentin-Anthony:
This PR makes two small changes:
- Sometimes the `python megatron/fused-kernels/setup.py install` will fail with errors that `pip install megatron/fused-kernels` will not cause (e.g. a nasty undefined torch symbol that I encountered today). I therefore updated the error message to recommend `pip install` instead.
- Failing to import a fused kernel can be due to many issues (i.e. broken install) besides the kernels not being installed, so we should print the specific error as well.


configs/neox_arguments.md
```diff
@@ -1845,11 +1847,3 @@ Args for deepspeed runner (deepspeed.launcher.runner).
 
     Adds a `--comment` to the DeepSpeed launch command. In DeeperSpeed this is passed on to the SlurmLauncher as well. Sometime necessary for cluster rules, or so I've heard.
 
-
-
-- **no_ssh_check**: bool
```
StellaAthena, line 1850: There was no mention of this being removed in the PR. Just want to check that it's deliberate / record the motivation.

Quentin-Anthony: It was done automatically in https://github.com/EleutherAI/gpt-neox/pull/824/commits/577126cf7c8532c7c8c29b04e173603987a37410 looks like. I'll take a look later as to why.

Quentin-Anthony: This will be resolved by https://github.com/EleutherAI/gpt-neox/pull/831

StellaAthena (approved):

PR #841, opened 2023-03-18
Support for LLaMA

zphang:


megatron/model/transformer.py
```diff
@@ -723,24 +792,33 @@ def forward(self, x, attention_mask, layer_past=None):
                 attention_output, presents = attention_output
                 self.layer_past = presents
             with torch.enable_grad():
-                attention_output = bias_dropout_fn(
-                    attention_output,
-                    bias=attention_bias.expand_as(residual),
-                    residual=residual,
-                    prob=self.hidden_dropout,
-                )
+                if attention_bias is not None:
```
StellaAthena, line 795: This code is a bit opaque to me. Is this check supposed to be for whether or not we have bias weights? I think we need to have some comments explaining it.

megatron/model/transformer.py
```diff
@@ -723,24 +792,33 @@ def forward(self, x, attention_mask, layer_past=None):
...
             )
+
             with torch.enable_grad():
-                output = bias_dropout_fn(
-                    mlp_output,
-                    bias=mlp_bias.expand_as(attention_output),
-                    residual=attention_output,
-                    prob=self.hidden_dropout,
-                )
+                if self.mlp_type == "llama":
```
StellaAthena, line 811: I would rather call this "no bias MLPs" than "llama MLPs." That will make it much easier to follow this code for anyone who hasn't read the Llama paper (recently).

megatron/neox_arguments/neox_args.py
```diff
@@ -343,6 +343,26 @@ class NeoXArgsModel(NeoXArgsTemplate):
...
+    use_bias_in_attn_linear: bool = True
+    """
+    If false, attn_linear (e.g. QKVO) will not have bias terms
+    """
+
+    mlp_type: str = "regular"
+    """
+    Types:
+        regular: Megatron implementation
+        llama: LLaMA MLP
```
StellaAthena, line 359: See prior comments about name change.

megatron/neox_arguments/neox_args.py
```diff
@@ -343,6 +343,26 @@ class NeoXArgsModel(NeoXArgsTemplate):
...
+    """
+    llama_mlp_multiple_of: int = 256
+    """
+    LLaMA MLP sizes are padded to a nice number
```
StellaAthena, line 363: Is this only applicable to llama MLPs? Can we add this functionality to the codebase as a whole?

StellaAthena (changes requested):

megatron/model/transformer.py
zphang: In addition to not having bias terms, the architecture is also different (rather then up-proj then down-proj, there's an additional gating linear map). I don't feel strongly about the naming either way

megatron/neox_arguments/neox_args.py
zphang: Actually, I might take this out. This is a constraint for how they use to compute the intermediate dimension, but we can just hard-code it into the configuration for easier readability.

StellaAthena: Okay, I think that makes sense then. So we’d have a “llama” config file which satisfies this, but not tell the user to follow it in general?

zphang:
Other than the above note on LLaMAMLP, I can incorporate the necessary changes and update the PR.

megatron/model/transformer.py
StellaAthena: With the different architecture, I now like keeping the LLaMA-specific name.

StellaAthena:
I'm confused by the `"make_vocab_size_divisible_by"` argument in the config file. Can you explain what it does? We generally assume that the data has been preprocessed including pre-tokenzied, and that's the point in time where it would make sense to have such a thing (and we do).

Otherwise, this looks good to me. @Quentin-Anthony, any thoughts?

StellaAthena:
recheck

Quentin-Anthony:
> I'm confused by the `"make_vocab_size_divisible_by"` argument in the config file. Can you explain what it does? We generally assume that the data has been preprocessed including pre-tokenzied, and that's the point in time where it would make sense to have such a thing (and we do).
> 
> Otherwise, this looks good to me. @Quentin-Anthony, any thoughts?

Looking good to me. I'm gonna run some tests then merge.

Quentin-Anthony:
Works for all of my tests. Merging.

Quentin-Anthony (approved):

DaoD:
@zphang Thanks for your work! It seems that there is no params.json in the llama checkpoint. Where can I get it? Thanks!

wiio12:
@zphang Hi, thank you for your work! can you provide a `readme` explaining how to run llama in neox?

wiio12:
> @zphang Thanks for your work! It seems that there is no params.json in the llama checkpoint. Where can I get it? Thanks!

Hi @DaoD, this `params.json` file can be found in the [offical checkpoint](https://github.com/facebookresearch/llama) (It requires to fill the google form).

wiio12:
Hi @DaoD, I think llama should be loaded with the `llama/7B.yml“ config file.

I post an [issue](https://github.com/EleutherAI/gpt-neox/issues/921) that discusses problems that I came across when loading llama in Neox.

DaoD:
@zphang I think we need a new convert_sequential_to_hf.py to convert the obatined model into HF style. Have you done something about this? Thanks!

borghives:
@zphang in the tools/convert_raw_llama_weights_to_neox.py, how does one convert llama tokenizer?  

- I have tried using the documented flag in the tool "--model_size tokenizer_only" (error "assert model_size in NUM_SHARDS" when trying to run the convert_raw_llama_weights_to_neox.py)
- I have tried using the huggingface llama tokenizer.json file (error missing "merge-file" when trying to run generate)

DaoD:
> @zphang in the tools/convert_raw_llama_weights_to_neox.py, how does one convert llama tokenizer?
> 
> * I have tried using the documented flag in the tool "--model_size tokenizer_only" (got error "assert model_size in NUM_SHARDS" when trying to run the convert_raw_llama_weights_to_neox.py)
> * I have tried using the huggingface llama tokenizer.json file (got error missing "merge-file" when trying to run generate)

I think you do not need to convert llama tokenizer. Just set
```
"tokenizer_type": "SPMTokenizer",
"vocab-file": "/llama-7b-hf/tokenizer.model",
```
The tokenizer.model can be obtained from this [link](https://huggingface.co/decapoda-research/llama-7b-hf/tree/main).

DaoD:
I find another problem. The eod token's id of SentencePieceTokenizer (eos_token_id=0) is differnet from the orginal LlamaTokenizer (eos_token_id =1), which may cause some problems in training and inference.

wiio12:
> I find another problem. The eod token's id of SentencePieceTokenizer (eos_token_id=0) is differnet from the orginal LlamaTokenizer (eos_token_id =1), which may cause some problems in training and inference.

Hi @DaoD, have you tried to do inference with this model, can you generate reasonable text with it?

DaoD:
> > I find another problem. The eod token's id of SentencePieceTokenizer (eos_token_id=0) is differnet from the orginal LlamaTokenizer (eos_token_id =1), which may cause some problems in training and inference.
> 
> Hi @DaoD, have you tried to do inference with this model, can you generate reasonable text with it?

Yes, but I do not use the inference code provided by gpt-neox. I just convert the model into Huggingface style, and use the HF function for generation. It seems correct.

wiio12:
> > > I find another problem. The eod token's id of SentencePieceTokenizer (eos_token_id=0) is differnet from the orginal LlamaTokenizer (eos_token_id =1), which may cause some problems in training and inference.
> > 
> > 
> > Hi @DaoD, have you tried to do inference with this model, can you generate reasonable text with it?
> 
> Yes, but I do not use the inference code provided by gpt-neox. I just convert the model into Huggingface style, and use the HF function for generation. It seems correct.

Thx! This helps a lot:)

sxthunder:
> 

Hello, can you convert a gpt-neox llama model to HF style?

PR #857, opened 2023-03-26
Rename pythia configs to represent the total parameter count

Quentin-Anthony:
- [ ] Create `configs/pythia` folder and populate with cleaned pythia configs


StellaAthena:
If we’re going to continue to add config files for new models we train, perhaps we should name them `pythia-70m.yml` or `pythia/70m.yml`? If only because it seems highly likely that we’ll train two different models with the same number of parameters at some point.

Quentin-Anthony:
> If we’re going to continue to add config files for new models we train, perhaps we should name them `pythia-70m.yml` or `pythia/70m.yml`? If only because it seems highly likely that we’ll train two different models with the same number of parameters at some point.

I'm fine with this, since it would also solve the issue of having a single `125M.yml` config, but it being unclear whether that's a GPT-style config or a Pythia-style config. I'll add a `pythia` folder under `configs` containing cleaned versions of the pythia configs (i.e. versions without cluster-specific paths and such).

Quentin-Anthony:
Resolved by https://github.com/EleutherAI/gpt-neox/pull/881

PR #862, opened 2023-03-29
add flash_attn_kvpacked

satpalsr:
**Issue**
Inference with flash attention gives error due to improper shapes of qkv, as key and value are cached.
To Reproduce:
1. Take any config with flash attention.
2. Run `python deepy.py generate.py configs/70M-deduped.yml -i input_prompt.txt -o prompt_out.txt` with some text in input_prompt.txt
3. Gives error due to different sq and sk values [here](https://github.com/EleutherAI/gpt-neox/blob/5561ff759c43c98fbd7927830b1a6f4eee74f979/megatron/model/transformer.py#L444)

This PR separates query from packed qkv matrix to resolve the issue.

### DIFF
--- a/megatron/model/transformer.py
+++ b/megatron/model/transformer.py
@@ -279,9 +279,16 @@ class ParallelSelfAttention(nn.Module):
             if self.use_flash_attention:
                 from megatron.model.flash_attention import (
                     flash_attn_unpadded_qkvpacked_func,
+                    flash_attn_unpadded_kvpacked_func,
                 )
 
-                self.flash_attention_function = flash_attn_unpadded_qkvpacked_func
+                self.flash_attn_unpadded_kvpacked_func = (
+                    flash_attn_unpadded_kvpacked_func
+                )
+                self.flash_attn_unpadded_qkvpacked_func = (
+                    flash_attn_unpadded_qkvpacked_func
+                )
+
                 if self.pos_emb == "alibi":
                     raise ValueError(
                         "Flash attention is currently not compatible with AliBi positional embeddings. Use sinuisoidal, learned, or rotary embeddings instead."
@@ -420,7 +427,7 @@ class ParallelSelfAttention(nn.Module):
         context_layer = context_layer.view(*output_size)
         return context_layer
 
-    def flash_attention(self, query_layer, key_layer, value_layer):
+    def flash_attention(self, query_layer, key_layer, value_layer, layer_past):
         # [b, np, sq, sk]
         output_size = (
             query_layer.size(1),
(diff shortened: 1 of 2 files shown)


DaoD:
I find this commit is similar to my issue (https://github.com/EleutherAI/gpt-neox/issues/883).
The general idea to tackle the problem is the same.
I think we do not need to add a new parameter layer_past in the function of flash_attention, but simply use `if self.training == True` to check whether it is training or inference stage.
High five for the same solution!!!

satpalsr:
@dashstander removed conflicts. Please test out now.
Thanks for the suggestion @DaoD

megatron/model/transformer.py
```diff
@@ -429,59 +435,101 @@ def flash_attention(self, query_layer, key_layer, value_layer):
...
+
+                # [sq, b, np, hn] -> [b * sq, np, hn]
+                query_layer = query_layer.transpose(0, 1).reshape(
+                    output_size[0] * output_size[2], output_size[1], -1
+                )
+
+                # Combined k/v into [b * sk, 2, np, hn].
+                kv = torch.concat([key_layer, value_layer], dim=1)
+
+                output = self.flash_attn_unpadded_kvpacked_func(
```
dashstander, line 477: this needs to be `self.flash_attention_function`

satpalsr: yeah. Did you attempted doing inference? If not, I'd do it and push if any more changes.

satpalsr: Will run the inference and push any changes as needed.

dashstander:
There are a few outstanding bugs that I have pointed out, though the general gist is ensuring that this works for both training and inference and that the code switches between those modes correctly. Some of that is actually not an issue with this PR and seems to be an issue with the `generate.py` script, as was mention in the (now closed) #883. Actively working on this now, should be able to iron this stuff out.

dashstander:
Just had a productive conversation with @satpalsr . Currently the implementation relies on the `torch.nn.Module` attribute `training` which isn't used elsewhere in the code base. There's also an issue with the three different FlashAttention functions getting assigned to the same name at initialization. Since models are always initialized with `self.training == True` even resetting it by calling `inference_mode` won't change the FlashAttention function.

StellaAthena:
> Since models are always initialized with self.training == True even resetting it by calling inference_mode won't change the FlashAttention function.

Where does this happen? I was expecting it to be in `setup_for_inference_or_eval` but it doesn't seem to be.

dashstander:
> Where does this happen? I was expecting it to be in `setup_for_inference_or_eval` but it doesn't seem to be.

[It's an attribute built in to `torch.nn.Module`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module) and just set to `True` by default. Part of the changes if we want to use this particular attribute will be setting it in the `training_mode` and `inference_mode` methods. An alternative might be to specifically add an inference vs training configuration to `NeoxArgs`, but that would definitely need to be part of a larger discussion.

dashstander:
I also just realized that this doesn't add the [Triton kvpacked function ](https://github.com/HazyResearch/flash-attention/blob/df1344f8668829d6e46b32d79d9089d19888da9f/flash_attn/flash_attn_triton.py#L794), so technically inference with Triton FlashAttention + ALiBi would be broken.

dashstander:
I tested and confirmed this all worked with [these changes.](https://github.com/EleutherAI/gpt-neox/files/11267413/flash_attn.patch)

dashstander (approved):
@satpalsr added me as a collaborator on their fork so I made the changes I was requesting. Looks good to me from here.

StellaAthena (approved):
Dash has tested and approved this code and I don't see anything that stands out as problematic.

PR #868, opened 2023-04-02
Update transformer.py

crazyofapple:
bug fixed. if not, the hidden size can not divided.


StellaAthena (approved):
Tested and verified to run. Math checks out per the geglu paper

PR #869, opened 2023-04-02
No longer ignore `--iteration` when passed to train.py

haileyschoelkopf:
We currently resume training using the checkpoint tag provided in the "latest" file within the save folder, following Deepspeed's default behavior. However, we want to be able to override this "latest" tag when launching a job when, for example, resuming several checkpoints back from a corrupted run.

An `--iteration` flag that can be passed to `deepy.py` already exists, and can be successfully used to evaluate a model from any step with evaluate.py , ignoring the "latest" file. However, though this flag properly sets `neox_args.iteration` at initialization, when initializing a model we ignore this because it is not passed to `setup_model_and_optimizer`.

This PR resolves this, allowing users to successfully use `--iteration` alongside `train.py`.

@Quentin-Anthony @ShivanshuPurohit

### DIFF
--- a/megatron/training.py
+++ b/megatron/training.py
@@ -190,7 +190,7 @@ def pretrain(neox_args):
     # Model, optimizer, and learning rate.
     timers("model and optimizer").start()
     model, optimizer, lr_scheduler = setup_model_and_optimizer(
-        neox_args=neox_args, use_cache=False
+        neox_args=neox_args, use_cache=False, iteration=neox_args.iteration
     )
     timers("model and optimizer").stop()
 


StellaAthena (dismissed):
Tested and found to work, but didn’t stress test unusual cases or potential human errors. Approving, but I’ll leave it to @Quentin-Anthony to decide if it should be merged or tested further.

Also, I noticed we are passing `use_cache=False`… do we never want to use cache?

haileyschoelkopf:
re: `use_cache=False`: yes, in training we don't ever want or need cache. If use_cache is ever true it gets set to false when enabling train mode: https://github.com/EleutherAI/gpt-neox/blob/038b01148f8faccdddb25f82ebc18e6706943ffe/megatron/model/gpt2_model.py#L331

Quentin-Anthony:
I'm happy with this. Merging.

PR #877, opened 2023-04-09
Make conversion scripts compatible with Python 3.8

### LINKED ISSUE #875: 1.0 HF conversion script fails on Python 3.8
**Describe the bug**

While the official recommended Python version for NeoX is 3.8, **convert_v1.0_to_hf.py** only works with Python 3.9 or later due to minor incompatibility (tested on Python 3.8.10).

>TypeError: 'type' object is not subscriptable

**Suggested change**

Modify `list[torch.Tensor]` at the line 42 to `List[torch.Tensor]` (capitalized L)
Add import `from typing import List`

haileyschoelkopf:
Closes #875 , and also improves error handling on FP16 detection for config YAMLs being converted.


tools/convert_sequential_to_hf.py
```diff
@@ -157,7 +157,16 @@ def convert(input_checkpoint_path, loaded_config, output_checkpoint_path):
 
     hf_model = GPTNeoXForCausalLM(
         hf_config
-    ).half()  # nice-to-have: lazy init weights somehow?
+    )
+
+    # save model in FP16 if Deepspeed fp16 was used in config, else 32 bit
+    fp16 = get_key(loaded_config, "fp16")
```
StellaAthena, line 163: Should there be logic here to handle bf16 explicitly? Even if it’s not mechanically necessary it’s probably going to be clearer to future users and maintainers.

haileyschoelkopf: Good call, will fix!

haileyschoelkopf:
PR updated and tested for different cases! 

Note that the way we specified precision in DeeperSpeed v1.0 diverges from the upstream Deepspeed docs, I think( can look further though?), and so these two scripts diverge at this conditional.

Quentin-Anthony:
I'm happy with this. Merging.

PR #900, opened 2023-04-23
fix requirements by using git+ and apply precommit

bzantium:
- It needs to be 'git+' to download from repository.
- Apply pre-commit

### DIFF
--- a/configs/neox_arguments.md
+++ b/configs/neox_arguments.md
@@ -926,7 +926,7 @@ Text Generation arguments
 
 - **prompt_end**: str
 
-    Default = 
+    Default =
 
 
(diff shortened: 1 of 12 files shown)


StellaAthena:
Thanks for noticing! I’ve already corrected this though, so if you can confirm that the current main works for you I’ll close the PR.

bzantium:
> Thanks for noticing! I’ve already corrected this though, so if you can confirm that the current main works for you I’ll close the PR.

Oh you just fixed it! Sure, it is ok to close this PR :)

PsuedoJack (commented):
{[(Retrieve and expose)]}://run again test international waters:)/

PR #907, opened 2023-04-27
Adds a script to convert HF checkpoints to NeoX 2.0 with mp and pp sharding

bentherien:
The "convert_hf_to_sequential.py" script allows users to convert model checkpoints from HF format to NeoX 2.0 and cache them in the specified (by the config file) pipe-parallel-size and model-parallel-size.

The main use case is to enable simple caching of pre-trained models for fine-tuning in NeoX 2.0. 

Included functionality:
    - convert_hf_to_sequential() conversion function tested with the Pythia suite
    - sharding for (MP=1, PP=0),  (MP=1,PP=1),  (MP>1,PP=0),  (MP>1,PP>1)
    - logit testing for comparing with HF models (only available for world_size=1)

Playing with the script and logit testing uncovers some interesting findings:
    1. Disabling flash attention for the converted Pythia 70M model leads to numerical overflow in NeoX 2.0.
    2. Even with flash attention enabled, the forward pass between the NeoX 2.0 model and the HF model is not identical (we found this to be caused by a difference in applying the rotary embedding).


tools/convert_hf_to_sequential.py
```diff
@@ -0,0 +1,490 @@
...
+        if reduce(np.logical_or,[x in k for x in ['layernorm', 
+                                                  'rotary_emb',
+                                                  'dense_4h_to_h.bias',
+                                                  'norm.weight',
+                                                  'norm.bias',
+                                                  'attention.dense.bias']]):
+            # no splitting
+            for x in range(num_mp_ranks):
+                ranks[x][k] = v
+        else:
```
Ayushk4, line 123: I think here ColumnParallel and RowParallel layers would have to be handled separately:

For example linear layer for query_key_value is difference from the dense in attention
https://github.com/EleutherAI/gpt-neox/blob/586f5149ea948cc46b270ddb9e8076628868c34a/megatron/model/transformer.py#L221

https://github.com/EleutherAI/gpt-neox/blob/586f5149ea948cc46b270ddb9e8076628868c34a/megatron/model/transformer.py#L305

Same goes for the two linear layers in MLP in each transformer layer.
https://github.com/EleutherAI/gpt-neox/blob/586f5149ea948cc46b270ddb9e8076628868c34a/megatron/model/transformer.py#L96
https://github.com/EleutherAI/gpt-neox/blob/586f5149ea948cc46b270ddb9e8076628868c34a/megatron/model/transformer.py#L106


As reference: The conversion scripts from 1.0 and current version to HF are handling it separately:

RowParallel Linear Layers:
https://github.com/EleutherAI/gpt-neox/blob/586f5149ea948cc46b270ddb9e8076628868c34a/tools/convert_sequential_to_hf.py#L188-L194

ColumnParallel Linear Layers:
https://github.com/EleutherAI/gpt-neox/blob/586f5149ea948cc46b270ddb9e8076628868c34a/tools/convert_sequential_to_hf.py#L208-L215

bentherien: So currently my script handles column and row parallel by assuming that the larger dimension is the one to be shared. Since the non-sharded dimension is typically the hidden state, this could cause problems for larger models. Let me fix this

StellaAthena:
@bentherien thank you for your contribution! Please sign the CLA and we will review and merge this :)

PR #936, opened 2023-05-14
remove password based login for root

AI-WAIFU:
Removes password based login for root, this should fix the security issues we've been having.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,12 +37,7 @@ RUN apt-get update -y && \
     pip install gpustat
 
 ### SSH
-# Set password
-ENV PASSWORD=password
 RUN mkdir /var/run/sshd && \
-    echo "root:${PASSWORD}" | chpasswd && \
-    # Allow root login with password
-    sed -i 's/#PermitRootLogin prohibit-password/PermitRootLogin yes/' /etc/ssh/sshd_config && \
     # Prevent user being kicked off after login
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -37,12 +37,7 @@ RUN apt-get update -y && \
...
 ### SSH
-# Set password
-ENV PASSWORD=password
 RUN mkdir /var/run/sshd && \
-    echo "root:${PASSWORD}" | chpasswd && \
-    # Allow root login with password
-    sed -i 's/#PermitRootLogin prohibit-password/PermitRootLogin yes/' /etc/ssh/sshd_config && \
     # Prevent user being kicked off after login
     sed -i 's@session\s*required\s*pam_loginuid.so@session optional pam_loginuid.so@g' /etc/pam.d/sshd && \
     echo 'AuthorizedKeysFile     .ssh/authorized_keys' >> /etc/ssh/sshd_config && \
```
ShivanshuPurohit, line 43: `echo 'PasswordAuthentication yes' >> /etc/ssh/sshd_config && \`
should this stay?

ShivanshuPurohit (approved):

Dockerfile
AI-WAIFU: Yes, because right now our current infra assumes that you can log in with a password. I need to change that before removing this line.

Issue #885, opened 2023-04-14
Unable to run generate text


hax4usincupad:
**Describe the bug**
 ```
File "generate.py", line 91, in <module>
    main()
  File "generate.py", line 33, in main
    model, neox_args = setup_for_inference_or_eval(use_cache=True)
  File "/home/ubuntu/gpt-neox/megatron/utils.py", line 443, in setup_for_inference_or_eval
    model, _, _ = setup_model_and_optimizer(
  File "/home/ubuntu/gpt-neox/megatron/training.py", line 621, in setup_model_and_optimizer
    model, optimizer, _, lr_scheduler = deepspeed.initialize(
  File "/home/ubuntu/.neox3.8/lib/python3.8/site-packages/deepspeed/__init__.py", line 137, in initialize
    assert config is None, "Not sure how to proceed, we were given deepspeed configs in the deepspeed arguments and deepspeed.initialize() function call"
AssertionError: Not sure how to proceed, we were given deepspeed configs in the deepspeed arguments and deepspeed.initialize() function call
```

**To Reproduce**
Just execute
` ./deepy.py generate.py -d configs bf16_125M.yml local_setup.yml`

**Environment (please complete the following information):**
 - GPUs: 4
- Configs:

StellaAthena:
@hax4usincupad where did you get model weights to run with this c? Did you train your own model in bf16?

hiroshi-matsuda-rit:
Try to downgrade the deepspeed to version 0.8.x. @hax4usincupad 
`pip install -U "deepspeed<=0.9.0"`

StellaAthena:
> Try downgrading the deepspeed to version 0.8.x. @hax4usincupad
> `pip install -U "deepspeed<0.9.0"`

This library is designed to use a custom fork of DeepSpeed: https://github.com/EleutherAI/DeeperSpeed. Using the `main` DS may cause issues

garrett361:
> This library is designed to use a custom fork of DeepSpeed...

@StellaAthena I'm hitting the same issue as OP and I'm a little confused about the gpt-neox/DeeperSpeed connection: where in the gpt-neox code base is DeeperSpeed being used specifically?

In particular, I see very few references to DeeperSpeed in the code base and `requirements/requirements.txt` seems to just pip install the usual `deepspeed`. I'm unable to see how DeeperSpeed makes it way into the Docker image, which is how I'm trying to run gpt-neox. Can you clarify?  Thanks!

[...]

PR #938, opened 2023-05-16
Remove duplicate deepspeed config and allow forced multinode

Quentin-Anthony:
We currently pass the `--deepspeed_config` in 
https://github.com/EleutherAI/gpt-neox/blob/056e9caf23e04a2ee13b5066cb6b7bef5404aa33/megatron/neox_arguments/arguments.py#L546
but also pass the config in
https://github.com/EleutherAI/gpt-neox/blob/056e9caf23e04a2ee13b5066cb6b7bef5404aa33/megatron/training.py#L627

This used to be allowed by DS, but the latest version requires us to choose ([DS commit](https://github.com/microsoft/DeepSpeed/commit/47f9f13bd3236d426d558d61ef358b36191ef026) and [gpt-neox user facing this issue](https://github.com/EleutherAI/gpt-neox/issues/885)). Restricting to a single config passed is backwards-compatible and fixes this issue.

Further, DeepSpeed will automatically launch with `torch.dist` if a single node is used, and will disregard if a user requests `srun` or `mpirun`. There are cases where a user needs `srun` or `mpirun` even on a single node, so this PR introduces `force_multi`, which exposes DS's internal param to force the requested multinode runner on a single node.

### DIFF
--- a/configs/neox_arguments.md
+++ b/configs/neox_arguments.md
@@ -111,7 +111,7 @@ Logging Arguments
 
 - **git_hash**: str
 
-    Default = 6e60b1e
+    Default = b130d58
 
     current git hash of repository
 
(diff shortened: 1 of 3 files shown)


Quentin-Anthony:
@StellaAthena -- Can you review/merge?

Issue #939, opened 2023-05-17
Running with bf16 error


Life-0-1:
**Describe the bug**
I tested bf16 training with various configurations, including the one given at `configs/bf16_125M.yml`, but all of them failed. 
I list the configurations and corresponding errors bellow.
Code is running on 8 A100 GPUS and the latest gpt-neox repo is used.

### Config 1, which is the same as `configs/bf16_125M.yml`
```
"zero_optimization": {
    "stage": 0,
    "allgather_partitions": True,
    "allgather_bucket_size": 500000000,
    "overlap_comm": True,
    "reduce_scatter": True,
    "reduce_bucket_size": 500000000,
    "contiguous_gradients": True,
  },

"fp16": {
     "enabled": true,
     "type": "bfloat16", # set bf16 as precision
     "loss_scale": 0,
     "loss_scale_window": 1000,
     "hysteresis": 2,
     "min_loss_scale": 1
   },
"fp32_allreduce": True, # without a patch to torch, bf16 models have to do the allreduce in fp32
```
**error**
```
Traceback (most recent call last):
  File "train.py", line 27, in <module>
    pretrain(neox_args=neox_args)
  File "gpt-neox/megatron/training.py", line 226, in pretrain
    iteration = train(
  File "gpt-neox/megatron/training.py", line 778, in train
    loss_dict, skipped_iter = train_step(
  File "gpt-neox/megatron/training.py", line 684, in train_step
    reduced_loss = train_step_pipe(
  File "gpt-neox/megatron/training.py", line 734, in train_step_pipe
    loss = model.train_batch(data_iter=data_iterator)
  File "anaconda3/envs/neox/lib/python3.8/site-packages/deepspeed/runtime/pipe/engine.py", line 346, in train_batch
    self._exec_schedule(sched)
  File "anaconda3/envs/neox/lib/python3.8/site-packages/deepspeed/runtime/pipe/engine.py", line 1374, in _exec_schedule
    self._exec_instr(**cmd.kwargs)
  File "anaconda3/envs/neox/lib/python3.8/site-packages/deepspeed/runtime/pipe/engine.py", line 658, in _exec_forward_pass
    outputs = super().forward(inputs)
  File "anaconda3/envs/neox/lib/python3.8/site-packages/deepspeed/utils/nvtx.py", line 11, in wrapped_fn
    ret_val = func(*args, **kwargs)
[...]

PR #941, opened 2023-05-17
Updates bf16 demo config and mixed precision docutmentation.

dashstander:
As was pointed out in #939 , the documentation and example configs no longer match NeoX's behavior regarding bf16 training. This updates `configs/bf16_125M.yml` and `configs/README.md` to be in sync with current functionality.


dashstander:
Also needed to disable the "repeated overflows" checks that NeoX's [OverFlowMonitor](https://github.com/EleutherAI/gpt-neox/blob/1a43a58d450aa4df07b1adf4f5acd9cd86d27899/megatron/utils.py#L341) was doing.

Quentin-Anthony (approved):

PR #966, opened 2023-06-04
Use minimal softmax mask values based on the datatype

bentherien:
The current GPT-NeoX implementation uses a hardcoded value of -10000.0 mask values in its softmax function. While this works fine most of the time, it can become problematic if unnormalized softmax values computed by the model become very small. Therefore, it is always best practice to take the smallest possible (based on the datatype) numerical value for softmax masking. This change does exactly this, following the implementation at https://github.com/huggingface/transformers/blob/539e2281cd97c35ef4122757f26c88f44115fa94/src/transformers/models/gpt_neox/modeling_gpt_neox.py#L226


megatron/model/gpt2_model.py
```diff
@@ -45,7 +45,11 @@
 
 
 def gpt2_attention_mask_func(attention_scores, ltor_mask):
-    attention_scores.masked_fill_(ltor_mask, -10000.0)
+    mask_value = torch.finfo(attention_scores.dtype).min
+    # Need to be a tensor, otherwise we get error: `RuntimeError: expected scalar type float but found double`.
+    # Need to be on the same device, otherwise `RuntimeError: ..., x and y to be on the same device`
+    mask_value = torch.tensor(mask_value, dtype=attention_scores.dtype).to(attention_scores.device)
```
Quentin-Anthony, line 51: It's always best to initialize tensors directly on the target device. Using `.to` necessitates a copy first on the torch default device (`'cpu'` unless explicitly set).

Change to:
`mask_value = torch.tensor(mask_value, dtype=attention_scores.dtype, device=attention_scores.device)`

bentherien: Sounds good, fixed below

PR #967, opened 2023-06-04
Updated the HF to Sequential script to be more general

bentherien:
Changes include
    1. ability to specify any pythia model for conversion
    2. bug fixes related to distributed caching
    3. HF model inference using the fast gelu function for better logit alignment


tools/convert_hf_to_sequential.py
```diff
@@ -379,45 +396,72 @@ def convert(hf_model, ckpt_dir, output_dir):
...
+        neox_args = NeoXArgs.from_ymls(args.config)
     neox_args.configure_distributed_args()
-    neox_args.build_tokenizer() 
+    neox_args.build_tokenizer()
+    neox_args.initialize_tensorboard_writer()
+
+
+    # setup logging and timers
+    init_wandb(neox_args=neox_args)
+    timers = Timers(
```
Quentin-Anthony, line 458: What's the point of this timer?

bentherien: The point is to mimic the training code to a T as that has been thoroughly tested

Quentin-Anthony: I understand the logic, but timing/logging is something we can safely ignore and I don't want wandb to be a dependency of this script.

Comment out lines 5, 23-26, and 457-460.

bentherien: Fair enough. I've commended those lines and also verified the script works on Summit.

Quentin-Anthony (approved):

PR #958, opened 2023-05-29
Adding the possibility of passing a label dataset

honglu2875:
As discussed with @haileyschoelkopf in DM, I'm cleaning up my personal fork and putting the major difference out here in case it is a useful feature. It allows to take an optional `label_data_paths` (which is assumed to be perfectly in sync with the training data). 

But if there is anything that does not fit with the current design or anything, feel free to close it though.

## Why do I need it
It was used when finetuning diff models, where a custom loss mask needs to be applied to the dataset. It is very controllable if I can do it on the dataset level and easy to double-check by poking into the generated masks.
I also used this in some experiments of finetuning pythia for text repairing. It works well.

## How it works
- Without the argument `label_data_paths`, nothing should change. 
- With the argument, it will 
  - also load the labels into training dataset (see `build_the_dataset` function), so that the dataset yields an extra "label" item;
  - apply loss mask to negative labels, shift by 1 and use it as target labels during training (see `_get_batch` function).

Besides the label, there is also a minor fix in `setup_for_inference_or_eval`. Not sure if it's an issue with newer version of DeepSpeed, but the `no_load_optim` wasn't enough to prevent the optimizer from loaded when I do inference and use stuff like `generate_samples_unconditional`.

PR #979, opened 2023-06-22
Dataload fix

jahatef:
In #958 

https://github.com/EleutherAI/gpt-neox/pull/979/files#diff-e2b248f8c422a601bcb0b7d93f96c1dff070f2694737e2b69f1def64ab9c1844L276

When `data = None`, `_keys = ["text", "label"]`

This led to access errors:

https://github.com/EleutherAI/gpt-neox/pull/979/files#diff-e2b248f8c422a601bcb0b7d93f96c1dff070f2694737e2b69f1def64ab9c1844R280-R287

where `data_b` does not have any values in the `"label"` key.

Instead, 

https://github.com/EleutherAI/gpt-neox/pull/979/files#diff-e2b248f8c422a601bcb0b7d93f96c1dff070f2694737e2b69f1def64ab9c1844R306

was placed to only access the `"label"` key if `neox_args.label_data_paths` is set from the config file, which is the config determining if a labeled dataset is used.

### DIFF
--- a/configs/1-3B.yml
+++ b/configs/1-3B.yml
@@ -3,7 +3,7 @@
    # parallelism settings ( you will want to change these based on your cluster setup, ideally scheduling pipeline stages
    # across the node boundaries )
    "pipe_parallel_size": 1,
-   "model_parallel_size": 2,
+   "model_parallel_size": 1,
 
    # model settings
    "num_layers": 24,
--- a/configs/1-3B.yml
+++ b/configs/1-3B.yml
@@ -74,7 +74,7 @@
    },
 
    # misc. training settings
-   "train_iters": 320,
+   "train_iters": 320000,
    "lr_decay_iters": 320000,
    "distributed_backend": "nccl",
    "lr_decay_style": "cosine",
--- a/configs/1-3B.yml
+++ b/configs/1-3B.yml
@@ -3,7 +3,7 @@
    # parallelism settings ( you will want to change these based on your cluster setup, ideally scheduling pipeline stages
    # across the node boundaries )
    "pipe_parallel_size": 1,
-   "model_parallel_size": 1,
+   "model_parallel_size": 2,
 
    # model settings
    "num_layers": 24,
@@ -74,7 +74,7 @@
    },
 
    # misc. training settings
-   "train_iters": 320000,
+   "train_iters": 320,
    "lr_decay_iters": 320000,
    "distributed_backend": "nccl",
    "lr_decay_style": "cosine",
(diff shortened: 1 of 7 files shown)


megatron/training.py
```diff
@@ -273,8 +273,7 @@ def pretrain(neox_args):
 
 def _get_batch(neox_args, tokenizer, keys, data, datatype):
     """Support function for get_batch / get_batch pipe (to avoid code repetition)"""
-    _keys = [k for k in keys if k in data] if data else keys
```
jahatef, line 276: if 'data  = None', this will set `keys = ["text", "label".

configs/1-3B.yml
```diff
@@ -3,7 +3,7 @@
    # parallelism settings ( you will want to change these based on your cluster setup, ideally scheduling pipeline stages
    # across the node boundaries )
    "pipe_parallel_size": 1,
-   "model_parallel_size": 1,
+   "model_parallel_size": 2,
```
Quentin-Anthony: Don't make config changes

data/enwik8/enwik8.zip
Quentin-Anthony, line 1: remove this change from the PR. We don't want `enwik8.zip` to be hosted on gh

megatron/data/data_utils.py
```diff
@@ -67,6 +67,7 @@ def build_the_dataset(
     indexed_dataset = make_indexed_dataset(data_prefix, data_impl, skip_warmup)
     if label_prefix is None:
         label_dataset = None
+
```
Quentin-Anthony, line 70: remove this space change

megatron/training.py
```diff
@@ -324,12 +323,12 @@ def get_batch(neox_args, data_iterator):
...
     """A modification of get_batch() to work with the latest batch instead of an iterator."""
     # Items and their type.
-    keys = ["text", "label"]
+    keys = ["text", "label"] if neox_args.label_data_paths else ["text"]
     datatype = torch.int64
 
     tokens, labels, loss_mask, attention_mask, position_ids = _get_batch(
         neox_args, neox_args.tokenizer, keys, data, datatype
-    )
+    )
```
Quentin-Anthony, line 331: remove this space from the pr

StellaAthena:
@Quentin-Anthony I have made the requested changes. I'm finding that training hangs in mulit-node settings with tp > 1, as reported in #985. However I see the same on `main` and don't think this PR is to blame.

StellaAthena (approved):
After applying this fix to _both nodes_ it now works. Whoops xD

PR #928, opened 2023-05-09
[Bug] Make Configs Consistent

austinburnett:
- #920 

**Checklist:**
- [x] I have added a warning that '-' may be deprecated in relevant readme.
- [x] I have updated all documentation to be consistent. Replacing all occurrences of '-' with '_'
- [x] I have modified demo files to also be consistent.

PR #986, opened 2023-07-04
Turn remaining hyphens into underscores in merge20b.py

wunderalbert:
I've noticed that gpt-neox has moved towards addressing config items using `_` rather than `-` to separate words (see e.g. #928).

The old way is still used in one script, and it bugged me... and it also seems risky: This script modifies fields, so if they already exist with `_` as should be expected, it will not overwrite as intended, but add additional fields with `-`, causing clashes down the line.

According to my regex (`\[['"][a-zA-Z]+-[a-zA-Z-]+['"]\]`), this is the only remaining occurrence of hyphens addressing config items in the code.

Also, this is my first contribution to gpt-neox. I did not find the contribution guidelines and apologise in advance if I'm trampling all over them; I'd be grateful if you let me know if I am.

### DIFF
--- a/tools/merge20b.py
+++ b/tools/merge20b.py
@@ -36,16 +36,16 @@ def modify_config(input_config_path, output_config_path, output_dir):
         loaded_config = yaml.full_load(f)
 
     # replace model/pipeline parallel
-    loaded_config["model-parallel-size"] = 1
-    loaded_config["pipe-parallel-size"] = 1
+    loaded_config["model_parallel_size"] = 1
+    loaded_config["pipe_parallel_size"] = 1
 
     # replace load / save directories:
     loaded_config["load"] = output_dir
     loaded_config["save"] = output_dir
 
     # replace some other paths
-    loaded_config["vocab-file"] = os.path.join(output_dir, "20B_tokenizer.json")
-    loaded_config["log-dir"] = "./logs"
+    loaded_config["vocab_file"] = os.path.join(output_dir, "20B_tokenizer.json")
+    loaded_config["log_dir"] = "./logs"
 
     # we need to make sure the resulting vocab size is correct
     # do this by modifying the 'make_vocab_size_divisible_by' argument to be


StellaAthena:
Thank you for flagging this for our attention. Most of the codebase actually continues to accept both for the sake of backwards compatibility, and I didn’t test the specific function you mention when looking at the previous PR. I want to inspect this carefully before proceeding.

What tests have you run to validate the behavior of your patch?

re: contributions guidelines… yeah we should definitely write some :( Unfortunately it’s not something we’ve had the bandwidth to do and most of the major contributors to this library have minimal experience with writing something like that… nothing seems objectionable in your PR to me, though admittedly I don’t lead this project.

wunderalbert:
> What tests have you run to validate the behavior of your patch?

Good question, imperfect answer: I haven't run it in actual "anger" on a model, since I do not have a model trained where this script applied (for one, my models have different key names, for another, this appears to be a blueprint script anyways -- you're supposed to modify it to your needs, and replace e.g. the hardcoded `global_step150000`). 

However, all changes in this PR are contained to the function `modify_config`, so I could just comment out the other functions (`modify model states` and `merge_model_weights`) from main and adjust the path.

The function takes the 20B.yml config (or rather the copy a model has made of it) and replaces some values. In the version of the script currently on `main`, running

```bash
python3 gpt-neox/tools/merge20b.py --input_dir checkpoints/ --output_dir test
```

yields a config file that has in it:

```
model-parallel-size: 1
model_parallel_size: 2
[...]
pipe-parallel-size: 1
pipe_parallel_size: 4
```

while the `head` version of this PR yields a config file that has only one copy of these similar parameters, 

```
model_parallel_size: 1
[...]
pipe_parallel_size: 1
```

StellaAthena:
@Quentin-Anthony I tested this and it seems to work.

Quentin-Anthony (approved):

PR #996, opened 2023-07-17
Replaced all torch.concat with torch.cat

mariebiscuit:
* Received "AttributeError: module 'torch' has no attribute 'concat'" on torch 1.8.1+cu111
* `torch.concat` is an alias for `torch.cat` and there are four other instances of `torch.cat `already being used in this file


StellaAthena (approved):
I double checked the documentation and ran this with the version that reported an error.

PR #1001, opened 2023-07-25
Add RoPE scaling 

ZHAOTING:
This PR basically syncs gpt-neox with [this HF PR #24653](https://github.com/huggingface/transformers/pull/24653).

Linear scaling and dynamic NTK scaling are provided as an alternative to the original RoPE.
It enables extending a pre-trained model's max input length (e.g. 2k -> 8k) by fine-tuning on longer sequences for only a few steps. More descriptions can be found in the HF PR.

Modifications:
* Modifications in `megatron/model/positional_embeddings.py` are almost identical to the HF PR and provide two new RoPE classes in addition to the original one.
* A new argument `rope_scaling` is added in `megatron/neox_arguments/neox_args.py` to select the wanted RoPE. It defaults to `rope_scaling=None` and corresponds to the original RoPE.
* The different rotary embedding classes are now loaded in `megatron/model/transformer.py` according to the selected scaling strategy.
* The check of`self.max_position_embeddings >= self.seq_length` in `megatron/neox_arguments/arguments.py` now takes into account the RoPE scaling options.

### DIFF
--- a/megatron/model/positional_embeddings.py
+++ b/megatron/model/positional_embeddings.py
@@ -36,31 +36,76 @@ class SinusoidalPositionalEmbedding(torch.nn.Module):
 
 
 class RotaryEmbedding(torch.nn.Module):
-    def __init__(self, dim, base=10000, precision=torch.half):
+    def __init__(self, dim, max_position_embeddings, base=10000, precision=torch.half):
         super().__init__()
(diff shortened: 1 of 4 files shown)


StellaAthena:
Thank you for submitting this PR. We’ve been in close contact with (and help fund) the researchers who produced this work. On their advice we’re going to hold off on implementing this for now, as there’s something new and exciting coming out soon that will change the HF implementation :)

PR #1035, opened 2023-09-18
integrated flash attention 2

a663E-36z1120:
Integrated flash attention 2 (ver. 0.2.2 -> 2.2.1). Wall-clock performance improvement observed for 125M param model with seq length 4096 and batch size 128 running on 1 a10 GPU.
image

### DIFF
--- a/megatron/model/flash_attention.py
+++ b/megatron/model/flash_attention.py
@@ -5,7 +5,7 @@ import torch.nn as nn
 import torch.nn.functional as F
 
 from flash_attn import flash_attn_triton
-import flash_attn_cuda
+import flash_attn_2_cuda as flash_attn_cuda # For flash_attn version 2.1.1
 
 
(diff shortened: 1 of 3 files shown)


megatron/model/flash_attention.py
```diff
@@ -5,7 +5,7 @@
 import torch.nn.functional as F
 
 from flash_attn import flash_attn_triton
-import flash_attn_cuda
+import flash_attn_2_cuda as flash_attn_cuda # For flash_attn version 2.1.1
```
Quentin-Anthony, line 8: Is this backwards-compatible with old flash_attn installations or will this break environments?

Quentin-Anthony: Ok it's not backwards-compatible. We'll just drop support for old flash attn and announce it.

Quentin-Anthony (approved):

PR #1039, opened 2023-09-22
Fixed final value of linear decay lr

foggy-frost-forest:
I found that the current linear decay scheduler is incorrect. When `self.num_iters` reached `self.end_iter`, the learning rate `lr` is not 0


Quentin-Anthony (approved):

Quentin-Anthony:
Thanks for catching this! This PR also made me realize the exponential lr decay needs the same fix. I'll add that PR as well shortly.

PR #1065, opened 2023-10-30
Edge-casing for multi-GPU HF-to-NeoX conversion

haileyschoelkopf:
This PR fixes a few bugs encountered when attempting to convert from HF to NeoX checkpoint format, via the invocation of the script through `deepy.py` in the multi-GPU case. 

After fixing these, and using `"launcher": "slurm", "deepspeed_slurm": true` in the config file I passed, conversion works for this distributed case. The distributed case can be always used if desired, even on a single GPU, and whether or not MP/PP are > 1.

### DIFF
--- a/tools/ckpts/convert_hf_to_sequential.py
+++ b/tools/ckpts/convert_hf_to_sequential.py
@@ -57,7 +57,7 @@ MULTI_GPU_ARGS = " ".join(
         "--output-dir checkpoints/neox_converted/pythia/70m",
         "--cache-dir checkpoints/HF",
         "--config configs/pythia/70M.yml configs/local_setup.yml",
-        "--test", 
+        "--test",
     ]
 )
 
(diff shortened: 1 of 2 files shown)


tools/ckpts/convert_hf_to_sequential.py
```diff
@@ -519,15 +519,15 @@ def get_non_existing_dir(tmp_dir):
     model, optimizer, _, lr_scheduler = deepspeed.initialize(
         model=model,
         optimizer=optimizer,
-        args=neox_args,
+        # args=neox_args,
```
Quentin-Anthony, line 522: Why was this done?

AIproj: An issue I had if you kept both args and config_params that led me to the same change, is that DeepSpeed would throw an error about being passed two sets of configs in `DeeperSpeed/deepspeed/__init__.py` : 
```
    if hasattr(args, "deepspeed_config") and args.deepspeed_config is not None:
        assert config is None, "Not sure how to proceed, we were given deepspeed configs in the deepspeed arguments and deepspeed.initialize() function call"
```

Quentin-Anthony: Ah yes this is a requirement in newer DeepSpeed. We can take this in then.

Quentin-Anthony (approved):

PR #1072, opened 2023-11-07
Fixing convert neox to huggingface bug

kyuheejang:
When processing mlp.dense_4h_to_h.bias and attention.dense.bias, tp_ranks are not reflected, so strange results always appear when tp_ranks is greater than 1.

When changing the code, we confirmed that the hugging face model (model_parallel_size 4 was applied) was converted to a neox model using convert_hf_to_sequential.py and converted to a huggingface model using convert_sequential_to_hf.py, making it the same as the original model.


Quentin-Anthony (approved):
LGTM. Will merge if @haileyschoelkopf approves.

haileyschoelkopf (approved):
LGTM!

PR #1073, opened 2023-11-07
Implement neox_args processing when OMPI_COMM_WORLD_SIZE>1

kyuheejang:
Looking at the code, there were parts that were not yet implemented, so I wrote a workable code.

Since args_parsed.megatron_config is delivered in base64 format, it must be converted before use.

### DIFF
--- a/tools/ckpts/convert_hf_to_sequential.py
+++ b/tools/ckpts/convert_hf_to_sequential.py
@@ -7,6 +7,8 @@ import deepspeed
 
 import argparse
 import torch
+import base64
+import json
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


tools/ckpts/convert_hf_to_sequential.py
```diff
@@ -496,7 +506,7 @@ def get_non_existing_dir(tmp_dir):
         # time.sleep(5)
 
     if int(os.environ.get("OMPI_COMM_WORLD_SIZE", 1)) > 1:
-        neox_args = consume_neox_args2(args2)
+        neox_args = consume_neox_args2(args)
```
StellaAthena, line 509: I don't think this is supposed to be changed to `args`.

PR #1074, opened 2023-11-08
Resolve existing CPU unit test errors

### LINKED ISSUE #1025: CPU Tests CI task is failing
Right now the [CPU tests are failing](https://github.com/EleutherAI/gpt-neox/actions/runs/6177402069/job/16768444935). Either the tests don't reflect the code or the code has a bug. Regardless they should be reconciled so that we can rely on CI for baseline quality checks.

mkerin:
Fixes #1025 

`test_neoxargs_usage` does a regex search for patterns of the form "args.*" and verifies that such used args are defined in `NeoXArgs`. In this case it picks up a false positive from 
https://github.com/EleutherAI/gpt-neox/blob/f214358baea68c1631e650e073371010b4b9b60e/megatron/neox_arguments/arguments.py#L938-L940

### DIFF
--- a/tests/neox_args/test_neoxargs_usage.py
+++ b/tests/neox_args/test_neoxargs_usage.py
@@ -51,6 +51,7 @@ def test_neoxargs_usage():
             "build_tokenizer",
             "attention_config[i]",
             "print",
+            "update"
         ]
     )
 


Quentin-Anthony:
Great to hear this was a simple fix

Quentin-Anthony (approved):

PR #1080, opened 2023-11-15
Extend ci suite

### LINKED ISSUE #957: Robust testing suite
Recently we’ve been having issues with new features breaking old ones. We do not currently have a detailed testing suite that validates all of the features of GPT-NeoX, but the codebase is large and complicated enough that we definitely should. This is going to be the start of a list of tests we should support off the top of my head, feel free to post comments with additional ones and I’ll add them. I’ll also start going through things and building it up systematically.

We have some tests [here](https://github.com/EleutherAI/gpt-neox/tree/main/tests/model) but they are missing notable new features and don’t seem to be updated as the library changes.

**Data Processing**
- [ ] Download scripts run
- [ ] Preprocessing with each supported tokenizer works
- [ ] Training a new tokenizer

**Primary Functions**
- [ ] Training (on one GPU, one node, and one pod)
- [ ] Inference
- [ ] Evaluation

**Optimizations and Parallelizations**
- [ ] ZeRO works and memory usage is within prescribed limits
- [ ] fp16 and bf16 work correctly
- [ ] Various MP and PP values work correctly
- [ ] 

**Conversion Scripts**
- [ ] NeoX -> HF library
- [ ] NeoX -> Megatron-DS
- [ ] NeoX -> SafeTensors

**Misc Features**
- [ ] Library installs correctly and packages don’t have conflicts
- [ ] MuP

mkerin:
Will hopefully close #957 when done.

**Breakdown of requested new unit test coverage**
```
| Description                                                             | Has Test | Test Function                       | Comment                                                                       |
|-------------------------------------------------------------------------|----------|-------------------------------------|-------------------------------------------------------------------------------|
| Data Processing                                                         | NA       | SECTION-HEADER                      |                                                                               |
| Download   scripts run                                                  | y        | test_url_accessibility              | Unavailable subsets   of ThePile are marked as skipped                        |
| Preprocessing   with each supported tokenizer works                     | y        | test_preprocess_data                | 5/6 tokenizers pass,   SentPiece failing                                      |
| Training a new   tokenizer                                              | y        | test_train_tokenizer                | Passes                                                                        |
| Primary   Functions                                                     | NA       | SECTION-HEADER                      |                                                                               |
| Launcher   scripts                                                      | y        | test_train_launcher                 | Passes                                                                        |
| Training (on   one GPU, one node, and one pod)                          | partial  | test_train_launcher                 | Only covers running on 1 node                                                 |
| Finetuning   (especially loading and training without optimizer states) | y        | test_finetune                       | Passes                                                                        |
| Inference                                                               | y        | test_generate                       | Passes                                                                        |
| Evaluation                                                              | y        | test_evaluate                       | Passes                                                                        |
| Optimizations   and Parallelizations                                    | NA       | SECTION-HEADER                      |                                                                               |
| ZeRO works and   memory usage is within prescribed limits               | n        |                                     |                                                                               |
| fp16 and bf16                                                           | y        |        test_model_training_options  | bf16 failing -   possibly due to incompatible options in the unit test set up |
| Optimizer   types                                                       | n        |                                     |                                                                               |
| Various MP and   PP values                                              | n        |                                     |                                                                               |
| Flash   Attention                                                       | n        |                                     | Out of scope                                                                  |
| Model Options                                                           | NA       | SECTION-HEADER                      |                                                                               |
| GPT-J residual                                                          | y        |        test_model_training_options  |                                                                               |
| LLaMA MLP                                                               | y        |        test_model_training_options  |                                                                               |
| Positional   embeddings                                                 | y        |        test_model_training_options  |                                                                               |
| Sparse   attention                                                      | y        |        test_model_training_options  |                                                                               |
| Dropout and   Weight decay                                              | y        |        test_model_training_options  |                                                                               |
| Kernel fusions                                                          | n        |                                     |                                                                               |
| With / without   bias terms                                             | y        | test_model_options                  |                                                                               |
| Conversion   Scripts                                                    | NA       | SECTION-HEADER                      |                                                                               |
| NeoX -> HF transformers library                                         | y        |        test_gpt_neox_to_huggingface | Failing; unclear how   to resolve                                             |
| NeoX ->   Megatron-DS                                                   | n        |                                     | Script doesn't exist                                                          |
| NeoX ->   SafeTensors                                                   | n        |                                     | Script doesn't exist                                                          |
| NeoX V1 ->   NeoX V2                                                    | n        |                                     | Script doesn't exist                                                          |
| Misc Features                                                           | NA       | SECTION-HEADER                      |                                                                               |
| Library   installs correctly and packages don’t have conflicts          | partial  | test_dependencies.py                |                                                                               |
| MuP (currently   bugged, see  #956)                                     | n        |                                     | Out of scope                                                                  |
```


mkerin:
Sorry - all reviewers please feel free to remove yourselves - I meant to open this as a draft PR for now.

mkerin:
Clean CI run of CPU only tests is here: https://github.com/mkerin/gpt-neox/actions/runs/6954696176

Quentin-Anthony:
> We're still missing a test coverage for a couple of things in Stella's initial list, but I think this is worth merging. @Quentin-Anthony if you could take a look when you have time it would be greatly appreciated.
> 
> The major categories of test coverage that we're still missing are:
> 
> * I have only added test coverage for MP=1 and PP=1
> * Tests to check that all supported optimizers run
> * Tests for flash attention
> 
> I won't be online much over the next couple of weeks, but I intend to take another look at these when I get back.
> 
> Some other issues that I encountered whilst working on this which are worth flagging:
> 
> * Running tests in parallel with `pytest --forked tests` is currently broken. I fixed one blocker for this (CUDA being initialised too early), but hit another (the master_socket is hardcoded, & if it’s busy - eg. By the first test to run - then subsequent tests which expect it to be available will fail). Running tests in serial with `pytest tests` is fine though.
> * We have a test for ‘convert_sequential_to_hf.py’, but it’s failing for reasons that are unclear to me. I’ve marked it as failure expected & left some notes about the error.
> * I think we should avoid downloading data for the CPU CI run, and instead rely on (small) test data stored in tests/data/

Thanks for this work! I'll review over the next couple of days.

StellaAthena:
My original request `Training (on one GPU, one node, and one pod)` is a typo, I meant to write `Training (on one GPU, one node, and two nodes)`.

I believe @zphang has used our library with the `SentencePieceTokenizer` and may be able to identify why its failing.

mkerin:
Thanks @StellaAthena.

To clarify, I believe that training on one and two nodes correspondes to training with world_size=1 or world_size=2 (equivalent to one or two cores on a GPU). So to test the first case of training on one GPU, we want to set the host file such that world_size =n where n is all cores available on that GPU?

mkerin:
The `best-download` dependency has been updated
image

& I confirmed that installing from source is now broken
https://github.com/mkerin/gpt-neox/actions/runs/7030214909
but installing from PyPi is fixed (for this gh run I dropped the commit to install best-download from latest gh instead of pypi)
https://github.com/mkerin/gpt-neox/actions/runs/7030248294

I updated the PR to use pypi as the source for best-download.

megatron/tokenizer/train_tokenizer.py
```diff
@@ -78,11 +78,12 @@ def train_tokenizer(
     tokenizer.train_from_iterator(json_iterator(input_dir), trainer)
 
     # And Save it
-    tokenizer.save(save_path, pretty=True)
-    print(f"Tokenizer saved at {save_path}")
+    if save_path:
```
Quentin-Anthony, line 81: what's the rationale of adding this `if save_path` conditional?

Quentin-Anthony:
Can confirm this is all working for me. Great work!

megatron/tokenizer/train_tokenizer.py
mkerin: I was just trying to reduce the amount of IO in tests here.

More specifically, I set `--tokenizer_output_path ""` in `tests/unit/test_tokenizer.py` so that after training a new tokenizer we don't bother saving the output. The other option is to use a temporary output location so that the new model is saved & then cleaned up after the end of the unit test... if you would prefer the latter then that's totally fine by me.

PR #1087, opened 2023-12-01
[muP] Rework

lintangsutawika:


megatron/model/init_functions.py
```diff
@@ -16,41 +16,22 @@
...
 
-try:
-    import mup
-except ImportError:
-    pass
 
-
-def init_method_normal(sigma, use_mup_outer=False, mup_init_scale=1.0):
+def init_method_normal(sigma):
     """Init method based on N(0, sigma)."""
```
ndey96, line 21: shouldnt it be N(0, sigma^2)?

megatron/model/init_functions.py
```diff
@@ -103,105 +84,91 @@ def init_(tensor):
...
-            mup.init.xavier_uniform_(tensor)
+    def init_(tensor, mup_m_width=mup_m_width):
+        init_weight = torch.nn.init.xavier_uniform_(tensor)
+        if mup_m_width != 1:
             with torch.no_grad():
-                tensor.mul_(mup_init_scale)
-            return tensor
-        else:
-            return torch.nn.init.xavier_uniform_(tensor)
+                init_weight.div_(mup_m_width)
```
ndey96, line 95: Bug: Dividing a random variable by `mup_m_width` will scale its variance by `1/(mup_m_width**2)`. Instead you want to divide by `sqrt(mup_m_width)`

megatron/model/init_functions.py
```diff
@@ -103,105 +84,91 @@ def init_(tensor):
...
-            mup.init.xavier_normal_(tensor)
+    def init_(tensor, mup_m_width=mup_m_width):
+        init_weight = torch.nn.init.xavier_normal_(tensor)
+        if mup_m_width != 1:
             with torch.no_grad():
-                tensor.mul_(mup_init_scale)
-            return tensor
-        else:
-            return torch.nn.init.xavier_normal_(tensor)
+                init_weight.div_(mup_m_width)
```
ndey96, line 109: Potentially same bug as above

megatron/model/init_functions.py
```diff
@@ -103,105 +84,91 @@ def init_(tensor):
...
-            mup.init.normal_(tensor, mean=0.0, std=std)
+    def init_(tensor, mup_m_width=mup_m_width):
+        init_weight = torch.nn.init.normal_(tensor, mean=0.0, std=std)
+        if mup_m_width != 1:
             with torch.no_grad():
-                tensor.mul_(mup_init_scale)
-            return tensor
-        else:
-            return torch.nn.init.normal_(tensor, mean=0.0, std=std)
+                init_weight.div_(mup_m_width)
```
ndey96, line 124: Potentially same bug as above

megatron/model/init_functions.py
```diff
@@ -103,105 +84,91 @@ def init_(tensor):
...
+    def init_(tensor, mup_m_width=mup_m_width):
+        init_weight = torch.nn.init.normal_(tensor, mean=0.0, std=std)
+        if mup_m_width != 1:
             with torch.no_grad():
-                tensor.mul_(mup_init_scale)
-            return tensor
-        else:
-            return torch.nn.init.normal_(tensor, mean=0.0, std=std)
-
+                init_weight.div_(mup_m_width)
```
ndey96, line 137: Potentially same bug as above

megatron/model/init_functions.py
```diff
@@ -103,105 +84,91 @@ def init_(tensor):
...
-            import mup
-        except ModuleNotFoundError:
-            print("Please install mup https://github.com/microsoft/mup")
-            raise Exception
-
     def _get(name):
         if name == "normal":
             return init_method_normal(
-                args.init_method_std, args.use_mup, args.mup_init_scale
+                sigma=args.init_method_std/math.sqrt(args.mup_m_width)
```
ndey96, line 148: if you are going to define the muP init_std adjustment here then you need to remove all the `init_weight.div_` calls so you dont double adjust

megatron/model/transformer.py
```diff
@@ -971,6 +971,9 @@ def parallel_lm_logits(input_, word_embeddings_weight, parallel_output, bias=Non
     else:
         logits_parallel = F.linear(input_parallel, word_embeddings_weight, bias)
 
+    if args is not None and args.use_mup:
+        logits_parallel /= args.mup_m_width
```
ndey96, line 975: you may also want to multiply by some tunable scalar value here (like we did in BLTM)

megatron/model/word_embeddings.py
```diff
@@ -155,9 +154,10 @@ def forward(self, input_ids, position_ids, tokentype_ids=None):
         # Dropout.
         embeddings = self.embedding_dropout(embeddings)
 
+        # Y_emb = m_emb * embed(X)
         if self.use_mup:
             with torch.no_grad():
-                embeddings.mul_(self.mup_embedding_mult)
+                embeddings = torch.mul(embeddings, self.mup_m_emb)
```
ndey96, line 160: seems correct, where mup_m_emb is a tunable scalar

megatron/mpu/layers.py
```diff
@@ -547,8 +548,10 @@ def set_parallel_output(self, value: bool):
         )  # if gather_output is True, parallel output is False, so we set the opposite
 
     def forward(self, input_):
+
+        # Y_logits = W_unembed * X / m_width
         if self.use_mup and self.mup_rescale_parameters:
-            input_ /= self.width_mult()
+            input_ /= self.m_width
```
ndey96, line 554: This is unnecessary and should be removed. The only adjustments that need to be made are to the LR and init_std. This will apply an adjustment twice. I think its safe to remove all muP stuff from this file

megatron/mup_substitute.py
```diff
@@ -39,31 +41,51 @@ def _get_coord_data(
 ):
     df = []
 
+    def word_embedding_coord_check_hook(module, input, output):
+        with torch.no_grad():
+            word_embedding_act_abs_mean_list.append(output.abs().mean().item())
+
+    word_embedding_act_abs_mean_list = []
```
ndey96, line 48: i think this needs to be defined above the hook?

megatron/mup_substitute.py
```diff
@@ -39,31 +41,51 @@ def _get_coord_data(
 ):
```
ndey96, line 41: To implement the coord check heres what I did:

`megatron/training.py::train()`
image

`megatron/training.py::training_log()`
image

Then to run a coord check I just configure a 10 step training run where these scalars are being logged. To make the coord check graphs you can just parse your tensorboard logs 
cc @lintangsutawika @Quentin-Anthony

ndey96: Then in `arguments.py` I add this to enable coord check:
image

megatron/learning_rates.py
```diff
@@ -95,8 +97,8 @@ def step(self, step_num=None):
         self.num_iters = step_num
         new_lr = self.get_lr()
         for group in self.optimizer.param_groups:
-            if self.use_mup and "width_mult" in group:
-                group["lr"] = new_lr / group["width_mult"]
+            if self.use_mup and ("lr_adjust" in group) and group["lr_adjust"] is True:
+                group["lr"] = new_lr / self.mup_m_width
```
ndey96, line 101: 👍🏾

megatron/model/init_functions.py
lintangsutawika: I'm split on what the best interface would be. For `init_method_normal` and `scaled_init_method_normal` I opted to not scale in the function but for other init functions I did.

megatron/model/transformer.py
lintangsutawika: I thought that `mup_m_width` was the tunable scalar for the logits here?

megatron/mpu/layers.py
lintangsutawika: I intended to implement the formula `Y_logits = W_unembed * X / m_width` from this table, specifically `output_logits_multiplier`
image

megatron/model/transformer.py
lintangsutawika: Is this the part where Y_logits = W_unemb X/m_width?

megatron/mpu/layers.py
lintangsutawika: Nvm, I realized where I was wrong. `Output logits multiplier` refers to the output embedding layer and not the output of the intermediate layers. I've removed the scaling in `layers.py`

megatron/mpu/layers.py
```diff
@@ -735,7 +739,7 @@ def set_parallel_output(self, parallel_output: bool):
 
     def forward(self, input_):
         if self.use_mup and self.mup_rescale_parameters:
-            input_ /= self.width_mult()
+            input_ /= self.m_width
```
ndey96: also redundant

lintangsutawika: this should also be removed.

megatron/model/transformer.py
```diff
@@ -971,6 +969,9 @@ def parallel_lm_logits(input_, word_embeddings_weight, parallel_output, bias=Non
     else:
         logits_parallel = F.linear(input_parallel, word_embeddings_weight, bias)
 
+    if args is not None and args.use_mup:
+        logits_parallel /= args.mup_width_multiplier
+
```
lintangsutawika, line 974:
```
logits_parallel /= args.mup_width_multiplier
logits_parallel *= args.mup_output_multiplier 
```

megatron/model/init_functions.py
```diff
@@ -103,105 +84,91 @@ def init_(tensor):
...
-            import mup
-        except ModuleNotFoundError:
-            print("Please install mup https://github.com/microsoft/mup")
-            raise Exception
-
     def _get(name):
         if name == "normal":
             return init_method_normal(
-                args.init_method_std, args.use_mup, args.mup_init_scale
+                sigma=args.init_method_std/math.sqrt(args.mup_width_multiplier)
```
jthestness, line 148: During our call, we noted that this leads to a bug: Since the width multiplier is applied to all layers, this doesn't allow the embedding layer to be initialized differently from the transformer backbone layers (precisely: muP prescribes that layers who's input and output dimensions both scale with width need to have a sqrt(width) multiplying factor).

jthestness: Would recommend refactoring this code: Remove the muP width multiplier completely from the initialization methods code, and only take the initializer parameters in here (e.g., standard deviation). Then, when the initializers are used from various layers, adjust the initializer based on that particular layer's muP width adjustment requirements.

rezaarefi (approved):

PR #1101, opened 2023-12-22
Lm eval 0.4.0 support

haileyschoelkopf:
@Quentin-Anthony 

This PR resolves incompatibilities resulting from upgrading from lm_eval v0.3.0 to v0.4.0 . 

Currently we haven't pushed v0.4.0 to PyPI--will update this requirement as soon as I've done so.


StellaAthena:
@haileyschoelkopf Have you checked that it runs during the course of training without breaking things?

haileyschoelkopf:
I tested on a saved checkpoint—will also test training shortly!

haileyschoelkopf:
Improved and tested running during training.

eval_tasks/eval_adapter.py
```diff
@@ -406,7 +406,6 @@ def run_eval(
                 "winogrande",
                 "mathqa",
                 "pubmedqa",
-                "triviaqa",
```
StellaAthena: Why was triviaqa removed?

eval.py
Quentin-Anthony, line 1: Why rename to `eval.py`?

Quentin-Anthony: Ah -- [rename evaluate.py to avoid shadowing HF evaluate library](https://github.com/EleutherAI/gpt-neox/pull/1101/commits/3d4d9e97d269f8ce764200040dccbf23ed7f4357)

eval_tasks/eval_adapter.py
Quentin-Anthony: added

Quentin-Anthony (approved):

Issue #1077, opened 2023-11-11
Apply new fused rotary embedding


Quentin-Anthony:
This could really help us: https://github.com/NVIDIA/apex/pull/1746

Quentin-Anthony:
May also need https://github.com/NVIDIA/apex/pull/1750 applied in order to test FYI

StellaAthena:
This is naively ported but untested.

Quentin-Anthony:
> This is naively ported but untested.

Huh? Please clarify.

StellaAthena:
> > This is naively ported but untested.
> 
> Huh? Please clarify.

Sorry, that was a progress update. I've ported the code changes to GPT-NeoX but haven't had a chance to test them yet.

Quentin-Anthony:
> > > This is naively ported but untested.
> > 
> > 
> > Huh? Please clarify.
> 
> Sorry, that was a progress update. I've ported the code changes to GPT-NeoX but haven't had a chance to test them yet.

Gotcha.

Quentin-Anthony:
@StellaAthena -- Can you write up what you have into a draft PR? We can offload testing from you.

StellaAthena:
@Quentin-Anthony They're not ready for testing yet, but I can still open a draft PR if you'd like. Right now I'm in a place where I think I've copied over all of the core code but the kernels aren't building and I haven't been able to debug why that is yet.

I had hoped to make more progress on this last week, but got swamped with some other stuff. I'm happy to hand it off if someone else wants to take it over.

Quentin-Anthony:
> @Quentin-Anthony They're not ready for testing yet, but I can still open a draft PR if you'd like. Right now I'm in a place where I think I've copied over all of the core code but the kernels aren't building and I haven't been able to debug why that is yet.
> 
> I had hoped to make more progress on this last week, but got swamped with some other stuff. I'm happy to hand it off if someone else wants to take it over.

Yes please make a draft PR

yang:
@StellaAthena Still wanting help getting this to work / testing this?

StellaAthena:
> @StellaAthena Still wanting help getting this to work / testing this?

That would be excellent, thank you.

PR #1103, opened 2023-12-23
Fused Rotary Embeddings

StellaAthena:
This is the start of an implementation of fused rotary embeddings for GPT-NeoX https://github.com/EleutherAI/gpt-neox/issues/1077. I assumed that it would be pretty straight forward despite my lack of familiarity with CUDA but I've been disabused of that notion.

Currently the issue is that there are errors when trying to compile the cuda extension.


Quentin-Anthony:
handled by https://github.com/EleutherAI/gpt-neox/pull/1108

PR #1105, opened 2023-12-24
fused layernorm

yang:
- Add simple util for timings
- Add fused layernorm kernel from Megatron

### DIFF
--- a/megatron/devutil.py
+++ b/megatron/devutil.py
@@ -1,0 +1,51 @@
+import torch.cuda
+
+
+class Metric:
+    """
+    Dumb utility to collect and report average wall-time metrics.
+    """
+
+    def __init__(self, label):
+        self.label = label
+        self.measurements = []
+
+    def collect(self, measurement):
+        self.measurements.append(measurement)
+
+    def get_measurements(self):
+        return self.measurements[:]
+
+    def report(self):
+        print(
+            self.label,
+            torch.quantile(torch.tensor(self.measurements), torch.arange(10) / 10.0),
+        )
+
+
+def monitor_method_cuda_wall_times(metric, obj, methodname):
+    """
+    Measure timings for a method on an object or class.
+
+    For instance:
+
+    >>> metric = Metric('!LNORM')
+    >>> monitor_method_wall_times(metric, LayerNorm, 'forward')
+    """
+    oldmeth = getattr(obj, methodname)
+
+    start_event = torch.cuda.Event(enable_timing=True)
+    end_event = torch.cuda.Event(enable_timing=True)
+
+    def newmeth(*args, **kw):
+        start_event.record()
+        try:
+            return oldmeth(*args, **kw)
+        finally:
+            end_event.record()
+            torch.cuda.synchronize()
+            elapsed = start_event.elapsed_time(end_event)
+            metric.collect(elapsed)
+            metric.report()
+
+    setattr(obj, methodname, newmeth)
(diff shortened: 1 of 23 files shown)


megatron/core/utils.py
```diff
@@ -0,0 +1,60 @@
+from functools import reduce
```
Quentin-Anthony, line 1: We don't have `megatron/core`, so this code needs added to `megatron/utils.py` along with updated imports

megatron/devutil.py
```diff
@@ -0,0 +1,36 @@
...
+    >>> monitor_method_wall_times(metric, layernorm, 'forward')
+    """
+    oldmeth = obj[methodname]
+
+    def newmeth(*args, **kw):
+        start = time.time()
+        try:
+            return oldmeth(*args, **kw)
+        finally:
+            end = time.time()
```
Quentin-Anthony, line 31: Using time.time() is inaccurate for CUDA code, especially distributed code. Use torch cuda events like in https://github.com/EleutherAI/cookbook/tree/main/benchmarks/communication

megatron/model/norms.py
```diff
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 import torch
-from torch.nn import LayerNorm as LayerNorm
-
+# from torch.nn import LayerNorm as LayerNorm
```
Quentin-Anthony, line 16: We need to add a config arg for the fused layernorm, and turn it on by default. People may want to turn it off. We should treat the layernorm kernel the same as our other fused kernels: https://github.com/EleutherAI/gpt-neox/blob/98716ebb4818498fe118ddb07c62bfa94a48da2a/megatron/neox_arguments/neox_args.py#L243-L261

megatron/neox_arguments/neox_args.py
```diff
@@ -125,6 +125,11 @@ class NeoXArgsModel(NeoXArgsTemplate):
     Normalization layer to use. Choose from "layernorm", "rmsnorm", "scalenorm".
     """
 
+    layernorm_fusion: bool = True
```
Quentin-Anthony, line 128: Can we do the same config line addition that was done for fused rope? I suspect users will want to tune this knob.

Quentin-Anthony: LGTM otherwise!

jahatef:
Howdy,

Upon testing, we found no speedup and eventual slowdown when training a 1-3B model for 320 iterations. You can see my test [here](https://wandb.ai/jacobhatef/test_fused?workspace=user-hatef-4).

StellaAthena:
> Howdy,
> 
> Upon testing, we found no speedup and eventual slowdown when training a 1-3B model for 320 iterations. You can see my test [here](https://wandb.ai/jacobhatef/test_fused?workspace=user-hatef-4).

It looks like the baseline the one with anomalous behavior here. Not to be an ass, but did you double and triple check by running it multiple times that something weird didn't just happen by magic to cause the baseline to speed up?

jahatef:
@StellaAthena I agree, so I ran some extra tests, still available from the above link. I see no difference between fused layer norm being on/off within the variance of the runs.

Quentin-Anthony:
I looked over these results with @jahatef and agree it falls within variance. I'm going to run some CUDA profiling to check if data movement is decreased with the layernorm kernel, and will merge if so. This would be on the grounds that as newer GPUs spend progressively less time on GEMMs, data movement becomes more critical.

Quentin-Anthony:
I see meaningful decrease in data movement and confirmed the preservation of accuracy. Merging. Thanks a ton for this @yang

Quentin-Anthony (approved):

Quentin-Anthony (approved):

PR #1108, opened 2024-01-02
Fused Rotary Embeddings (fixed)

yang:
Orig PR started by Stella:

> This is the start of an implementation of fused rotary embeddings for GPT-NeoX #1077. I assumed that it would be pretty straight forward despite my lack of familiarity with CUDA but I've been disabused of that notion.
> 
> Currently the issue is that there are errors when trying to compile the cuda extension.


StellaAthena:
Supercedes https://github.com/EleutherAI/gpt-neox/pull/1103

configs/125M_fused_rope.yml
```diff
@@ -0,0 +1,95 @@
...
+   "norm": "layernorm",
+   "pos_emb": "rotary",
+   "no_weight_tying": true,
+   "gpt_j_residual": false,
+   "output_layer_parallelism": "column",
+
+   # these should provide some speedup but takes a while to build, set to true if desired
+   "scaled_upper_triang_masked_softmax_fusion": false,
+   "bias_gelu_fusion": false,
+   "rope_fusion": true,
```
Quentin-Anthony, line 23: this should be added with a `false` default to all configs (leave pythia and llama configs alone) similar to the other fused kernels so that users get the suggestion.

This file also can be removed, accounting that the above has been added.

Quentin-Anthony (approved):

PR #1113, opened 2024-01-06
Add docker compose and change containerized setup instructions to use it

segyges:
Self-explanatory, probably. Quality of life improvement. Mostly because I was tired of typing out long docker commands.

I neglected to explain why to build locally instead of pulling from remote, which is because remote might be on a different commit, with different dependencies. Only way to guarantee matched versions is to run a script to check current commit and populate a variable with that when pulling, which seems excessive. I figure since it's not the primary way listed to do it, anyone doing that knows what they are doing and matching the commit is for them to do.


segyges:
I should probably make separate "checkpoints" and "data" directories, actually, so let me see if I can get to that before this gets considered too seriously.

Quentin-Anthony:
Great work! I just brought down your docker config to only include the relevant changes (data/ckpt paths) to avoid config bloat and to make its inclusion clear. I also applied the pre-commit.

Quentin-Anthony (approved):

segyges:
@Quentin-Anthony I tried that and it didn't work but looking at the config now I can't see why that would be, I am guessing I messed something else up when doing it. I will test it from current main when I get the chance later, will probably try to add a separate readme for the repro in configs/docker. It turns out the pythia 19m will run on 2x 3090, by the way. ;)

PR #1125, opened 2024-01-19
Minor changes

segyges:
- Changes default output for docker container
- Renames docker pythia config to indicate it is docker pythia config
- Fix unit test for evaluate function


tests/README.md
```diff
@@ -15,19 +15,18 @@ python prepare_data.py
...
 * A subset of tests can be selected by pointing to the module within tests
 
 ```bash
-# run all tests, output coverage report of megatron module in terminal
-pytest --forked --cov-report term --cov=megatron tests
+# run all tests, output coverage and html report of megatron module in terminal
+pytest --cov-report term --cov=megatron --html=test-report.html --self-contained-html tests
 
-# run tests in tests/model, output coverage report of megatron module as html
-pytest --forked --cov-report html --cov=megatron tests/model
```
Quentin-Anthony, line 27: Why the removal of `--forked`? These should be necessary for distributed pytests.

segyges: --forked is causing this error, a lot:

```
RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method
```

Removing it resolves those; it's possible it would be better to figure out why the runner fails in this way presently, though.

segyges: @Quentin-Anthony I pulled the bit about --forked out, resolving whatever is going on with tests can be pushed to a different issue/PR.

Quentin-Anthony (approved):
Thanks for these!

Issue #479, opened 2021-12-12
Add Mixture of Experts


sdtblck:
from [DeepSpeed-MoE for NLG: Reducing the training cost of language models by 5 times ](https://www.deepspeed.ai/news/2021/12/09/deepspeed-moe-nlg.html).

It should be a fairly simple addition as [the codebase they open source](https://github.com/microsoft/Megatron-DeepSpeed/tree/moe-training) is largely similar to ours (same base model, although we have diverged a bit since).

PR #1129, opened 2024-01-23
Add MoE

yang:
Closes https://github.com/EleutherAI/gpt-neox/issues/479


Quentin-Anthony (approved):
Tested and working. Thanks @yang !

megatron/training.py
```diff
@@ -387,19 +387,27 @@ def forward_step(
...
     ):
         loss_mask = loss_mask[:, : neox_args.curriculum_seqlen].contiguous()
         labels = labels[:, : neox_args.curriculum_seqlen].contiguous()
-    loss = cross_entropy(
+    main_loss = cross_entropy(
         outputs, (labels, loss_mask), _fp16=neox_args.fp16_lm_cross_entropy
     )
     if neox_args.memory_profiling:
         torch.cuda.nvtx.range_pop()
+    moe_loss = neox_args.moe_loss_coeff * sum(m.item() for m in moe_losses)
```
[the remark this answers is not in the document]
yang: You're right

configs/125M-moe.yml
```diff
@@ -0,0 +1,99 @@
+# GPT-2 pretraining setup
+{
+   # Have 4 experts per layer (every 2 layers by default)
```
StellaAthena, line 3: Let's explicitly show people which layers are experts in this example config.

StellaAthena: Also, is there a specific reason to have it be "every x layers" instead of letting the user specify which layers it is precisely?

I think it makes sense to model this argument after `attention_config`.

Quentin-Anthony: I agree that this would be more flexible, but I advocate for splitting this "per-layer control" support into a separate PR. 

Now that we're adding support for RWKV/Mamba/MoE, I suspect more users will want to combine blocks in new ways (e.g. attn --> mamba --> MoE ...). Doing this all at once instead of having another MoE-specific per-block flag will save us work in the long run.

In short, I think this PR is fine as-is and per-layer control should be separated into another PR.

Quentin-Anthony (approved):

PR #1138, opened 2024-01-29
Contributing Guide

### LINKED ISSUE #1110: Add a Contributor Guide
We should document some of our contributor guidelines (pre-commit, tests, etc) now that we have them.

A good one to model after: https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md

jahatef:
fixes #1110

### DIFF
--- a/configs/neox_arguments.md
+++ b/configs/neox_arguments.md
@@ -111,7 +111,7 @@ Logging Arguments
 
 - **git_hash**: str
 
-    Default = 765c219
+    Default = ef54045
 
     current git hash of repository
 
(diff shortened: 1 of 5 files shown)


configs/neox_arguments.md
```diff
@@ -261,6 +261,14 @@ Model Arguments
 
 
 
+- **layernorm_fusion**: bool
```
Quentin-Anthony, line 264: why is this here?

jahatef: outstanding question. Will remove.

jahatef: Actually this belongs here. Just not in this commit. This is autogenerated by https://github.com/EleutherAI/gpt-neox/pull/1138/commits/e840a061611507f972f0d381eeaac48cc336b338. I think this is a results of our docs updater not working before

Quentin-Anthony (approved):

PR #1144, opened 2024-02-04
Remove gas (beano)

segyges:
per code being removed:

```gas: int = None
    """gradient_accumulation_steps"""  # TODO this is a duplicate, remove?
```

This is set in many configs and [never assigned from](https://github.com/search?q=repo%3AEleutherAI%2Fgpt-neox%20gas&type=code). Its continued existence in neox only encourages people to set ```gas```, which does apparently nothing, instead of setting ```gradient_accumulation_steps```, which is probably what they intended to do.

Pythia configs left alone because the source of truth for those configs is the pythia repo.

### DIFF
--- a/configs/19M.yml
+++ b/configs/19M.yml
@@ -44,7 +44,7 @@
   },
 
   "train_micro_batch_size_per_gpu": 4, #32,
-  "gas": 1,
+  "gradient_accumulation_steps": 1,
   "data_impl": "mmap",
   "num_workers": 1,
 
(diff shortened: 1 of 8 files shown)


segyges:
Oh: And any model run with "gas" set at anything other than "1" will have had batch size smaller by 1/their_gas_size than they intended. People might want to check their previous work for that.

Quentin-Anthony (approved):

Quentin-Anthony:
Good catch!

PR #1145, opened 2024-02-04
Copy pythia configs

segyges:
1) This copies pythia configs directly from the pythia repo, which should remove all possibility for configuration drift
2) This was a real problem -- e.g., flashattention configuration existed in pythia but not in the neox copy of the 1b pythia config
3) Copying all the configs over exactly prevents this from ever happening again, prevents me from having to diff the files one by one to see if they are actually identical or differ at some value in the future, and makes it easier to keep up to date because we can just run a fresh copy from the pythia repo.
4) It might have slight differences in formatting because the pre-commit hook ran over it before I disabled the pre-commit hook for the configs/pythia directory
5) The pre-commit hook errors on one of the configs for no reason I can discern; I think it is probably actually a glitch in their yaml parser but I am not inclined to try to fix the yaml parser in check-yaml.
6) Even if that config is incorrect, which I do not think it is, it's better to have the same error in both repos than a mystery error that only exists in one of them.

### DIFF
--- a/configs/pythia/1-4B.yml
+++ b/configs/pythia/1-4B.yml
@@ -1,23 +1,26 @@
 {
-  "pipe_parallel_size": 1,
-  "model_parallel_size": 1,
+  # parallelism
+  "pipe-parallel-size": 1,
+  "model-parallel-size": 1,
 
-  "num_layers": 24,
-  "hidden_size": 2048,
-  "num_attention_heads": 16,
-  "seq_length": 2048,
-  "max_position_embeddings": 2048,
-  "pos_emb": "rotary",
-  "rotary_pct": 0.25,
-  "no_weight_tying": true,
-  "gpt_j_residual": true,
-  "output_layer_parallelism": "column",
+  # model settings
+  "num-layers": 24,
+  "hidden-size": 2048,
+  "num-attention-heads": 16,
+  "seq-length": 2048,
+  "max-position-embeddings": 2048,
+  "pos-emb": "rotary",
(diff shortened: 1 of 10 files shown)


Quentin-Anthony:
A few comments:
- There shouldn't be hanging paths in configs, which is why I don't want configs to exactly match the Pythia repo's. This was done for all the existing configs, and it looks like I missed it in your last commit: https://github.com/EleutherAI/gpt-neox/commit/e6e944acdab75f9783c9b4b97eb15b17e0d9ee3e
- With that in mind, the deduped configs would all need dropped, because their only meaningful difference from the non-deduped configs is paths to deduped datasets
- That said, I'm not sure why the 1B flash was dropped and admit it was likely a mistake. 
- I'm not worried about maintaining the configs from the Pythia repo because they exist just as research artifacts and will not be changed. They can only be added to like the 31M and 14M, and those can be ported here once as needed.

Therefore, what I'd actually like here is the per-file diff + to remove the 31M/14M data paths, and we use the new configs as a source of truth. Ideally only the 1B was modified and that's all we change. They'll all be identical to the non-deduped Pythia configs, but with all data paths removed.

segyges:
@Quentin-Anthony I had to do something weird with git but the current file diff on this branch faithfully shows the differences between the pythia repository and this one.

I am actually convinced by this that this is a bad idea, so I'm closing the pr. I will leave it as-is for reference on the diff and open another one for anything that seems like it is important out of, ah, everything.

Issue #1152, opened 2024-02-18
Add PyTorch Memory Profiler


Quentin-Anthony:
https://zdevito.github.io/2022/12/09/memory-traces.html

PR #1153, opened 2024-02-21
Memory profiling

jahatef:
Fixes https://github.com/EleutherAI/gpt-neox/issues/1152


megatron/training.py
```diff
@@ -1033,3 +1080,11 @@ def evaluate_and_print_results(
     print_rank_0("-" * length)
     print_rank_0(string)
     print_rank_0("-" * length)
+
+def save_snapshot(neox_args):
+    snapshot = torch.cuda.memory._snapshot()
+    snapshot_path = os.path.join(neox_args.memory_profiling_path)
```
Quentin-Anthony, line 1086: add an assert to make sure `memory_profiling_path` is not `None`, which is the arg default

jahatef: added

Quentin-Anthony (approved):

PR #1156, opened 2024-02-26
Fused kernel support for AMD (using JIT)

### LINKED ISSUE #954: Officially Support AMD GPUs
Currently, AMD GPU support lives experimentally in https://github.com/EleutherAI/gpt-neox/compare/main...AMD

We should port the kernel guards in https://github.com/microsoft/Megatron-DeepSpeed/commit/b4d4a0e06577218a21abf18eb7d8088bb8ecaad3#diff-059209398b62b21e2524b387fc6fd23ded28ec725798322c266ee4246d253670 to gpt-neox and bring this into `main`

- [x] Test latest gpt-neox `main` on AMD GPUs with fused kernels set to `false`
- [x] Port fused kernels to AMD hardware and test them on AMD GPUs
- [x] Add conditional HIP guards so that the same fused kernel code can run on AMD and NVIDIA GPUs without modification
- [x] Add flash-attn fallbacks because 2.x is not yet supported there

R0n12:
# Overview
This PR aims to bring JIT support into building fused kernels. The main reason for the swtich is that hip extension acts weird (making ninja generates different rules for the same target) with the traditional `CUDAExtension` setup, thus in this PR we use `cpp_extension.load()`, this enables no code modification when building across AMD and NVIDIA platforms

# Heads-up
This PR also bring in the `cpp_extention_load_helper` from [here](https://github.com/microsoft/Megatron-DeepSpeed/commit/b4d4a0e06577218a21abf18eb7d8088bb8ecaad3l), note that we JIT load fused kernels on demand and depends on `neox_args`, this may introduce additional complexity in testing, for fused kernels testing, I recommend keeping the original `setup.py` file as a backup.

Aim to close  #954


R0n12:
Will start a new branch and PR with all these changes, current state looks kinda messy.

Quentin-Anthony:
Closing to make room for the clean PR

PR #1157, opened 2024-02-26
Add Mamba Architecture

### LINKED ISSUE #1148: Add basic Mamba block
We want to add Mamba to gpt-neox:

- [x] Add basic mamba block, without kernels, from https://github.com/state-spaces/mamba/tree/main/mamba_ssm/modules to https://github.com/EleutherAI/gpt-neox/tree/main/megatron/model
- [x] Add mamba kernels from https://github.com/state-spaces/mamba/tree/main/mamba_ssm/ops
- [ ] Add config options for mamba
- [x] Add assertions to gpt-neox so that parallelism schemes and other architectures are disabled when mamba is enabled in config

haileyschoelkopf:
closes #1148 

This PR adds Mamba to NeoX, along with flags for turning on/off the selective scan + conv1d + full mamba_inner_fn kernels.

For now, does not support parallelism, but want to investigate adding Tensor Parallel to this.


haileyschoelkopf:
This seems to train well without parallelism, but am having bugs in a conversion script I wrote (gibberish output). I'll be checking for differences in output between a single instantiated layer for this versus the `mamba_ssm` module to try to diagnose further

haileyschoelkopf:
Worked after https://github.com/state-spaces/mamba/pull/211 !

Got performance from a 160m (trained with Pythia config, untied embed + unembed) on-par with the Mamba-130m results in paper! image

Will cleanup code slightly and add sample configs, then mark ready for review. This also pairs with a DeeperSpeed PR I'll make that should allow for holding specified parameters in fp32 despite Deepspeed trying to cast everything to 16 bit.

I want to check out adding tensor parallelism for Mamba too, but will do that later.

Quentin-Anthony:
> Worked after [state-spaces/mamba#211](https://github.com/state-spaces/mamba/pull/211) !
> 
> Got performance from a 160m (trained with Pythia config, untied embed + unembed) on-par with the Mamba-130m results in paper! image
> 
> Will cleanup code slightly and add sample configs, then mark ready for review. This also pairs with a DeeperSpeed PR I'll make that should allow for holding specified parameters in fp32 despite Deepspeed trying to cast everything to 16 bit.
> 
> I want to check out adding tensor parallelism for Mamba too, but will do that later.

Awesome! Great work. I have some TP ideas that we can discuss on discord.

haileyschoelkopf:
Ready for initial review! 

Pairs with https://github.com/EleutherAI/DeeperSpeed/pull/61 , which I'd appreciate feedback on if the approach there is acceptable.

megatron/model/mamba/mamba.py
```diff
@@ -0,0 +1,360 @@
...
+        # in https://arxiv.org/pdf/2312.00752.pdf Algorithm 2
+        # (computes data-dependent B, C, Delta/dt)
+        self.x_proj = nn.Linear(
+            self.d_inner, self.dt_rank + self.d_state * 2, bias=False, **factory_kwargs
+        )
+        init_method(self.x_proj.weight)
+
+        # up-project dt / Delta from dt_rank to d_inner
+        self.dt_proj = nn.Linear(
+            self.dt_rank, self.d_inner, bias=True, **factory_kwargs
```
Quentin-Anthony, line 91: We hardcode the conv biases, but the other projection biases get `self.use_bias`.

Why is this different?

haileyschoelkopf: No good justification--I can add 2 neox arguments to control the presence of conv, and in/out_proj biases.

haileyschoelkopf: Updated!

Quentin-Anthony: I like it, but these two `nn.Linear` biases are still set to `False` and `True` regardless of the new args?

haileyschoelkopf: the Mamba codebase exposes options in the Mamba block for `bias: bool = False` (controls in_proj and out_proj bias presence, as done here now) and `conv_bias: bool = True` (controls bias presence in conv1d). The other `x_proj` and `dt_proj` biases (False and True, respectively) are hardcoded and there aren't flags for them.

`dt_proj` 's bias is a special case and I believe we should keep it turned on -- Alg. 2 in the Mamba paper defines Delta as `Delta = Tau_{Delta}(Parameter + s_{Delta}(x))` where `s_{Delta}(x) = Broadcast_{D}(Linear_{1}(x))` or as they further explain in section 3.6 can be also `s_{Delta}(x) = Linear_{D}(Linear_{R}(x))` where Linear_R is the delta portion of `x_proj` and Linear_D is the `dt_proj` weight. Then, the `Parameter` term from Alg. 2 can be viewed as the bias term in `dt_proj`, with a special initialization from https://arxiv.org/abs/2206.12037 . 

TL;DR: `dt_proj.bias` is the `Parameter` in Alg. 2 and its presence is assumed.

I'm not aware of anything mentioned in the paper about `x_proj` not being able to have a bias though, and I can't think of a reason it is required to not have it--the fused inner function also seems not to support a bias on it. https://github.com/state-spaces/mamba/blob/34076d664838588a3c97727b263478ab9f621a07/mamba_ssm/ops/selective_scan_interface.py#L164-L167

I've opened an issue to clarify if the hardcoded value is a mistake.

haileyschoelkopf: Made x_proj bias now also controlled by the same flag!

configs/neox_arguments.md
```diff
@@ -432,7 +432,7 @@ Model Arguments
     The first item in the list specifies the attention type(s), and should be a list of strings. The second item
     specifies the number of times to repeat those attention types in the full list.
 
-    attention type choices:  [global, local, sparse_fixed, sparse_variable, bslongformer, bigbird, "gmlp", "amlp", "flash"]
+    attention type choices:  [global, local, sparse_fixed, sparse_variable, bslongformer, bigbird, "gmlp", "amlp", "flash", "mamba"]
```
Quentin-Anthony, line 435: **Note for the future:** In addition to the `attention_config`, we should add a `block_config` where the user can choose on a per-block basis where to place individual blocks for MLP(and its variants like MoE)/Mamba/Attention(and its variants). The `attention_config` would then just allow the user to choose between attention variants for any attention blocks in the broader `block_config`. A similar `mlp_config` could be useful.

E.g.
```
"block_config": ["mamba", "attention", "mamba", "attention", "mlp"],
"attention_config": [[["flash"], 2]],
"mlp_config": [[["moe"], 1],
```

One drawback from this strategy is that we're adding a lot of annoying book-keeping for the user.

It's already a bit confusing that we're putting the attention-free `mamba` block under `attention_config`, but I think it's fine for this PR.

megatron/model/mamba/mamba.py
Quentin-Anthony: Added this as a comment in case future code-readers have this same confusion.

megatron/model/norms.py
```diff
@@ -57,6 +57,7 @@ def __init__(self, dim, p=-1.0, eps=1e-8, bias=False):
             self.register_parameter("offset", self.offset)
 
     def forward(self, x):
+        dtype = x.dtype
```
Quentin-Anthony, line 60: In the future, we should add support for the triton RMSNorm kernel introduced by mamba. Noting here and adding a TODO for later.

Quentin-Anthony: https://github.com/state-spaces/mamba/blob/v1.2.0/mamba_ssm/ops/triton/layernorm.py

Quentin-Anthony:
**Note for the future:** In addition to the `attention_config`, we should add a `block_config` where the user can choose on a per-block basis where to place individual blocks for MLP(and its variants like MoE)/Mamba/Attention(and its variants). The `attention_config` would then just allow the user to choose between attention variants for any attention blocks in the broader `block_config`. A similar `mlp_config` could be useful.

E.g.
```
"block_config": ["mamba", "attention", "mamba", "attention", "mlp"],
"attention_config": [[["flash"], 2]],
"mlp_config": [[["moe"], 1],
```

One drawback from this strategy is that we're adding a lot of annoying book-keeping for the user.

It's already a bit confusing that we're putting the attention-free `mamba` block under `attention_config`, but I think it's fine for this PR.

Quentin-Anthony:
In the future, we should add support for the triton RMSNorm kernel introduced by mamba. Noting here and adding a TODO for later.

https://github.com/state-spaces/mamba/blob/v1.2.0/mamba_ssm/ops/triton/layernorm.py

megatron/model/mamba/mamba.py
```diff
@@ -0,0 +1,378 @@
+import math
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+try:
+    from mamba_ssm.ops.selective_scan_interface import (
```
Quentin-Anthony, line 8: Needs a new requirements file. Adding now.

Quentin-Anthony (approved):

PR #1162, opened 2024-02-29
Improve argument validation for Flash-attn + SWA

haileyschoelkopf:
As raised by @malteos , Flash attention requires not 2.0.0+ but version 2.3.0+ to support sliding_window_width. However also, sliding-window attention is supported in the absence of Flash.

This PR changes the argument validation to 1) check for version 2.3.0 and 2) only perform this check if "flash" type attention is used. Also, it doesn't pass window_size kwarg when not required, meaning we'll never give the invalid kwarg to flash versions < 2.3.0 .

### DIFF
--- a/configs/mistral/7B.yml
+++ b/configs/mistral/7B.yml
@@ -11,7 +11,7 @@
   "num_kv_heads": 8,
   # per Mistral, Mistral-7B-v0.1 was pretrained with 8192 seqlen
   # and instruction tuned to 16384 seqlen, all with 4096 sliding window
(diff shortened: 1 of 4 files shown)


haileyschoelkopf:
Also updates Mistral config with seqlen info following https://discord.com/channels/1144547040454508606/1144547040928481394/1156616342682751087 .

Quentin-Anthony (approved):
LGTM

KidiIT (commented):
All reviewed and approved🫶❤️

PR #1170, opened 2024-03-04
Single node Pythia 14M training on ngc pytorch 24.02 container

tf-nv:
Swapped the raw cuda container for the ngc pytorch 24.02 container.

This one has OpenMPI, PyTorch and Apex already installed, so those setup steps can be removed.

`flash-attn` was recently hotfixed for that container version, so 2.5.6 is a hard requirement.

To validate the setups, I ran a few training steps of Pythia 14M on 8xA100 and a 22B model on 8xA100 and 8xH100 w/ TP8.

The apt and pip install lists and requirements files could be cleaned up. E.g. trition is already present and gets downgraded. I will leave it like this for now in order to keep this PR small.

For multi node tests I will have to adapt the launcher first.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM nvidia/cuda:12.1.1-devel-ubuntu22.04
+FROM nvcr.io/nvidia/pytorch:24.02-py3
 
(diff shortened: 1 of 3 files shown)


Quentin-Anthony:
Yep this LGTM. Agree on the requirements cleanup, but let's leave that for a followup.

Quentin-Anthony (approved):

Issue #1174, opened 2024-03-07
MoE loss variable not defined in gpt j residual code path


tf-nv:
Running pythia 14M on master:

```
File "/gpt-neox/train.py", line 34, in <module>
    main()
  File "/gpt-neox/train.py", line 30, in main
    pretrain(neox_args=neox_args)
  File "/gpt-neox/megatron/training.py", line 228, in pretrain
    iteration = train(
  File "/gpt-neox/megatron/training.py", line 913, in train
    loss_dict, skipped_iter = train_step(
  File "/gpt-neox/megatron/training.py", line 793, in train_step
    loss = forward_step(
  File "/gpt-neox/megatron/training.py", line 391, in forward_step
    maybe_tuple = model((tokens, position_ids, attention_mask), neox_args=neox_args)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/deepspeed/utils/nvtx.py", line 15, in wrapped_fn
    ret_val = func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/deepspeed/runtime/engine.py", line 1822, in forward
    loss = self.module(*inputs, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
  File "/gpt-neox/megatron/model/utils.py", line 190, in forward
    x = func(forward_input)
  File "/gpt-neox/megatron/model/utils.py", line 181, in exec_func
    inputs = layer(inputs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
[...]

PR #1175, opened 2024-03-07
Cleaner dockerfile: Remove already installed deps

tf-nv:
Cleaning up the Dockerfile after the ngc pytorch switch (#1170):

- Eliminate already installed apt packages
- sparse attn requirement lead to a triton downgrade
- flash attn is already part of the ngc container (in another version that is compatible with TE)

The flash-attn change might be required for TE, see the [TE setup script](https://github.com/NVIDIA/TransformerEngine/blob/b459ccc93549c13797f884271904e019ad7fa3db/setup.py#L268). The following error from the docker log was indicating so, too. I am not 100% sure if the version upper limit is a technological requirement, or just a precaution. In any case this is now resolved, plus there's no need to build flash-attn.

```
#14 10.98 Installing collected packages: flash-attn
#14 10.98   Attempting uninstall: flash-attn
#14 10.98     Found existing installation: flash-attn 2.4.2
#14 10.98     Uninstalling flash-attn-2.4.2:
#14 11.74       Successfully uninstalled flash-attn-2.4.2
#14 13.81 ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
#14 13.81 transformer-engine 1.3.0+5b90b7f requires flash-attn!=2.0.9,!=2.1.0,<=2.4.2,>=2.0.6, but you have flash-attn 2.5.6 which is incompatible.
#14 13.81 Successfully installed flash-attn-2.5.6
```

Training pythia14M works (with gpt-j-residual=false , see #1174)
Training a 20B model with TP8 works as well

Just a few steps on 8xH100, not to convergence :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,11 +26,11 @@ LABEL org.opencontainers.image.base.name="nvcr.io/nvidia/pytorch:24.02-py3"
 #### System package (uses default Python 3 version in Ubuntu 20.04)
 RUN apt-get update -y && \
     apt-get install -y \
-    git python3-dev libpython3-dev python3-pip sudo pdsh \
-    htop tmux zstd software-properties-common build-essential autotools-dev \
-    nfs-common pdsh cmake g++ gcc curl wget vim less unzip htop iftop iotop ca-certificates ssh \
-    rsync iputils-ping net-tools libcupti-dev libmlx4-1 infiniband-diags ibutils ibverbs-utils \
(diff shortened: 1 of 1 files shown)


Quentin-Anthony (approved):
LGTM. 

The requirements files are intended for source-builds anyway, so any package overlap between them and the Dockerfile should be handled like this PR (i.e. removing the requirements file from Dockerfile)

PR #1177, opened 2024-03-08
Remove unused requirements-sparseattention

segyges:
- Removes this requirements file
- Removes reference to it in the dockerfile
- Also removes protobuf from the dockerfile since it is apparently unused and I noticed it
- Causes confusion and possible problems while installing since it will force a downgrade of Triton
- If downgrading Triton is actually essential I am wrong and this PR should be rejected.


Quentin-Anthony:
Removing sparse attn from Dockerfile + removing the Dockerfile Flash mismatch was just handled in https://github.com/EleutherAI/gpt-neox/pull/1175, and as you mentioned we still need protobuf. 

I would, however, still be interested in just updating the version of Flash to be consistent with the Dockerfile + removing `requirements/requirements-sparseattention.txt`

PR #1178, opened 2024-03-08
Resolve issues between kv_cache and flash attention.

chaochen99:
When using kv cache and flash attention in conjunction, it's crucial to set the causal parameter of flash_varlen_qkv_fn to False. Failing to do so will lead to inaccurate results.


Quentin-Anthony (approved):
LGTM. @haileyschoelkopf -- Please sanity-check this as well.

haileyschoelkopf:
LGTM to me as well--reference: https://github.com/Dao-AILab/flash-attention/blob/9818f85fee29ac6b60c9214bce841f8109a18b1b/flash_attn/modules/mha.py#L504

haileyschoelkopf (approved):

PR #1179, opened 2024-03-08
remove `best_download` as dependency

haileyschoelkopf:
we no longer use best_download as of updating the lm-eval-harness integration, so removing it from requirements.


Quentin-Anthony (approved):

KidiIT (commented):
All reviewed and approved🫶❤️

PR #1184, opened 2024-03-12
Mamba + Tensor Parallel Support

haileyschoelkopf:
This PR adds Mamba + TP support.

Loss curves comparing TP=2 to TP=1 with + without `mamba_inner_func_fusion`:

image

Versus when allreduce was missing with inner func fusion turned on:

image

Also tested that PP seems to work.

### DIFF
--- a/configs/neox_arguments.md
+++ b/configs/neox_arguments.md
@@ -111,7 +111,7 @@ Logging Arguments
 
 - **git_hash**: str
 
-    Default = 33f2842
+    Default = 696454f
 
     current git hash of repository
 
--- a/configs/neox_arguments.md
+++ b/configs/neox_arguments.md
@@ -111,7 +111,7 @@ Logging Arguments
 
 - **git_hash**: str
 
-    Default = fdac107
+    Default = f70c54d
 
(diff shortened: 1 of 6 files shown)


Quentin-Anthony:
LGTM, no comments.

@haileyschoelkopf -- As a final check, I'd like to verify that TP gives the expected memory benefits. Can you link the wandb here so I can take a look?

haileyschoelkopf:
Yep! https://wandb.ai/eleutherai/mamba-neox-tp-memsavings/workspace?nw=nwuserschoelkopf

this should be public and a clean comparison, lmk if more is needed or for some reason it's not visible (MBS=16 for both TP=1 and TP=2, corrected for w/ grad accum.) Seeing 29-30GB for TP=1, 19GB for TP = 2. 

Here's the full wandb of trial runs including initial tests, Mamba-160m, and some Pythia-160m baseline curves: https://wandb.ai/eleutherai/mamba-neox?nw=nwuserschoelkopf

Quentin-Anthony (dismissed):

Quentin-Anthony (approved):

PR #1188, opened 2024-03-18
[AMD] Supporting fused kernels build using JIT

R0n12:
This PR aims to enable JIT compilation for fused kernels on AMD GPUs, so that the same code will run on both AMD and NVIDIA GPUs. 

Previous `python setup.py install` method has issues with hipifying files, which results in ninja generating multiple rules for the same target. 

This will also make the extra fused kernel building instruction in the readme unnecessary. I will take care of the readme parts once this get approved

- [x] builds on AMD (MI100, MI250X) and NVIDIA (A100)
- [x] `test_fused_kernels.py` passed

#954 
Reference: https://github.com/microsoft/Megatron-DeepSpeed/commit/b4d4a0e06577218a21abf18eb7d8088bb8ecaad3

### DIFF
--- a/megatron/fused_kernels/__init__.py
+++ b/megatron/fused_kernels/__init__.py
@@ -12,13 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
+
 import os
 import pathlib
 import subprocess
-
-from pathlib import Path
-
-srcpath = Path(__file__).parent.absolute()
+import torch
+from torch.utils import cpp_extension
 
 # Setting this param to a list has a problem of generating different
 # compilation commands (with different order of architectures) and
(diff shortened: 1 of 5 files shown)


megatron/fused_kernels/__init__.py
```diff
@@ -12,13 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
```
Quentin-Anthony, line 15: Why include this copyright?

StellaAthena: Unless this file is verbatim from NVIDIA's code, the copyread header should look more like [this](https://github.com/EleutherAI/gpt-neox/blob/9ddff09d360fa9e281c96df48ad1bf42c05eb5f1/megatron/training.py)

R0n12: This code is borrowed from 
https://github.com/microsoft/Megatron-DeepSpeed/blob/main/megatron/fused_kernels/__init__.py

I have pushed a correct format

StellaAthena: It's not quite the same wording as we use elsewhere, but the new header looks adequate.

StellaAthena:
@Quentin-Anthony should we update the README to reflect this change?

Quentin-Anthony:
> @Quentin-Anthony should we update the README to reflect this change?

Yes. We're on it now.

PR #1190, opened 2024-03-18
[ZeRO-3] Partitioned init with `deepspeed.zero.Init()`

### LINKED ISSUE #1189: Large model instantiation using `DeepSpeed.zero.Init` for extremely large model under ZeRO-3
**Is your feature request related to a problem? Please describe.**
Currently GPT-NeoX doesn't support partitioned model initialization when using ZeRO-3, which will cause OOM error in most cases.

**Describe the solution you'd like**
A simple fix like this will do the trick inside `get_model`

```
    if neox_args.zero_stage == 3:
        with deepspeed.zero.Init():
            model = GPT2ModelPipe(
                neox_args=neox_args,
                num_tokentypes=0,
                parallel_output=True,
                topology=mpu.get_topology(),
                use_cache=use_cache,
            )
``` 

**Describe alternatives you've considered**
Other things I have in mind is to figure out a way to properly test this, I have tested this on a 175B model and it works. Please let me know if there's other testing needed

**Additional context**
Related issue: https://github.com/huggingface/accelerate/issues/922

R0n12:
fixes #1189

### DIFF
--- a/megatron/training.py
+++ b/megatron/training.py
@@ -24,6 +24,7 @@ from functools import partial
 
 import math
 import sys
+from contextlib import nullcontext
 
 import torch
 import deepspeed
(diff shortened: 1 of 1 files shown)


Quentin-Anthony:
Thanks @R0n12! Can you quickly test if my cleanup changes still work for your case?

R0n12:
- [ ] 13B over 16 A100
- [ ] 6.7B over 16 A100

I think we are good!

Quentin-Anthony (approved):

PR #1198, opened 2024-03-31
add rwkv support

jahatef:
This is ready for review.


SmerkyG:
Just a heads up, you can probably remove support for rwkv_pre_ffn and rwkv_mishglu as they aren't really used in any production RWKV models.

jahatef:
done

megatron/model/rwkv/rwkv.py
```diff
@@ -0,0 +1,356 @@
...
+
+    def __init__(self, neox_args, layer_number):
+        super().__init__()
+        self.neox_args = neox_args
+        self.layer_number = layer_number
+        self.fp16 = neox_args.precision == "fp16"
+        self.bf16 = neox_args.precision == "bfloat16"
+        if not hasattr(neox_args, "dim_att"):
+            neox_args.dim_att = neox_args.hidden_size
+        if not hasattr(neox_args, "dim_ffn"):
```
SmerkyG, line 280: RWKV 5+ use a 3.5x FFN multiplier, not 4x. There is unfortunately some misleading code in the RWKV-LM repo that has led many to make this mistake.

StellaAthena: I think we should probably let people experiment with this ratio however they want though, right? We can provide scripts that train the official versions, but if someone wants to train with a different ratio let them (much like we do with transformers).

SmerkyG: Oh yeah it definitely needs to stay parameterizable by the end user. The only change I'm requesting is that the default be 3.5x - this is done to reduce the parameters taken up by the gate that was added in v5, to stay at equiparams, and is what is in every released RWKV-6.0 model and described in the paper.

As an aside, it looks like v6.1+ is going to remove the gate and this will go back up to a default of 4x at that time.

Quentin-Anthony: Added versioning dirnames and changed this to 3.5x for v6

Quentin-Anthony (approved):

PR #1201, opened 2024-04-22
Create cmake-multi-platform.yml

Romario242003:


Quentin-Anthony:
Thanks for the contribution @Romario242003!

Can you provide some context for this addition? What problem is it solving and what motivated its inclusion.

PR #1209, opened 2024-05-04
Fix bug in tools/ckpts/convert_neox_to_hf.py for setting intermediate_size

jvendrow:
In tools/ckpts/convert_neox_to_hf.py, for neox architecture the 'intermediate_size' argument is not explicitly set, so it defaults to 24576 from:

https://github.com/huggingface/transformers/blob/9fe3f585bb4ea29f209dc705d269fbe292e1128f/src/transformers/models/gpt_neox/configuration_gpt_neox.py#L48

Proposed solution: set intermediate-size to 4 * hidden-size.


Quentin-Anthony:
@haileyschoelkopf -- FYI

Quentin-Anthony (approved):

PR #1210, opened 2024-05-06
Dmoe integration

DayOfThePenguin:
Supersedes #1197

This PR adds dropless MoE support using the Grouped GEMM implementation in megablocks.

## Features

Unlike the legacy DeepSpeed MoE implementation that uses the data parallel groups for expert parallelism, this implementation uses the model parallel group to parallelize the experts.  This avoids the following problems:

- Using data parallel groups to distribute the experts will incur inter-node communications to do a forward pass through a single layer
- MoE + pipeline parallelism is very complicated to reason about when you have MoE weights distributed across data parallel groups & deepspeed doesn't natively support it.

Clarified arguments by separating MoE args into their own class.

Use sinkhorn routing by default, support k>=1. TopK routing is used for evaluation/inference.

## Testing

Tested PP [3, 2, 1] and MP [1, 2, 4, 8] on Ampere GPUs.

## Notes

Added megablocks and grouped_gemm to the dependencies. It might be desirable to pull some of the kernels in directly like in NVIDIA megatron-core.


megatron/mpu/mappings.py
```diff
@@ -185,9 +180,102 @@ def _dmoe_gather(input_: torch.Tensor, tokens_per_expert: torch.Tensor):
     # Note: torch.cat already creates a contiguous tensor.
     output = torch.cat(tensor_list, dim=gather_dim)
 
-    # Bf16 convert
```
aurelion-source: .

aurelion-source:
Profiles before and after the merge:
https://wandb.ai/shetano-personal/dmoe/reports/DMOE--VmlldzoxMDU0NDY0OQ

megatron/mpu/mappings.py
aurelion-source: This was resolved in the latest commit.

configs/125M-dmoe.yml
```diff
@@ -67,35 +58,26 @@
 
    # regularization
    "gradient_clipping": 1.0,
-   "weight_decay": 0.1,
+   "weight_decay": 0.0,
```
Quentin-Anthony, line 61: there appear to be a lot of extraneous config changes. Any reason why?

megatron/neox_arguments/arguments.py
```diff
@@ -1075,15 +1077,6 @@ def calculate_derived(self):
         # if we set pipe_parallel_size to 0, GPT2ModelPipe.to_sequential() is called, and we run training with
         # the sequential model without the PipelineModule wrapper to avoid the overhead it incurs
         self.update_value("is_pipe_parallel", self.pipe_parallel_size >= 1)
-        if self.moe_num_experts > 1:
```
Quentin-Anthony, line 1078: Did we test these parallelism combinations?

aurelion-source: All parallelism combinations are now supported except for zero stages 2 and 3 due to using bf16.
I added assertions accordingly.

configs/125M-dmoe.yml
aurelion-source: Updated the config file. It now matches the default.

Quentin-Anthony (approved):

chuanyang-Zheng:
If possible, could I know why Top-K cannot be used for training? It seems that there is no error.

PR #1212, opened 2024-05-10
Add `intermediate_size` to GPT-NeoX models

dtamayo-nlp:
The current implementation only allows to set `intermediate_size` for Llama models, but I would like to be capable to change the `intermediate_size` in GPT-NeoX models. 

I have tested this implementation under a quick training, inference and conversion and it seems that it doesn't give any bugs. I hope it helps!

### DIFF
--- a/configs/1-3B.yml
+++ b/configs/1-3B.yml
@@ -3,7 +3,7 @@
    # parallelism settings ( you will want to change these based on your cluster setup, ideally scheduling pipeline stages
    # across the node boundaries )
    "pipe_parallel_size": 1,
-   "model_parallel_size": 1,
+   "model_parallel_size": 2,
 
    # model settings
    "num_layers": 24,
@@ -48,6 +48,7 @@
     "reduce_bucket_size": 500000000,
     "contiguous_gradients": True,
   },
+  "expansion_factor": 8,
 
    # batch / data settings
    "train_micro_batch_size_per_gpu": 4,
@@ -76,8 +77,8 @@
    },
 
    # misc. training settings
-   "train_iters": 320000,
-   "lr_decay_iters": 320000,
+   "train_iters": 10,
+   "lr_decay_iters": 10,
    "distributed_backend": "nccl",
    "lr_decay_style": "cosine",
    "warmup": 0.01,
--- a/configs/1-3B.yml
+++ b/configs/1-3B.yml
@@ -3,7 +3,7 @@
    # parallelism settings ( you will want to change these based on your cluster setup, ideally scheduling pipeline stages
    # across the node boundaries )
    "pipe_parallel_size": 1,
-   "model_parallel_size": 2,
+   "model_parallel_size": 1,
 
    # model settings
    "num_layers": 24,
--- a/configs/1-3B.yml
+++ b/configs/1-3B.yml
@@ -48,7 +48,6 @@
     "reduce_bucket_size": 500000000,
     "contiguous_gradients": True,
   },
-  "expansion_factor": 8,
 
    # batch / data settings
    "train_micro_batch_size_per_gpu": 4,
@@ -77,8 +76,8 @@
    },
 
    # misc. training settings
-   "train_iters": 10,
-   "lr_decay_iters": 10,
+   "train_iters": 320000,
+   "lr_decay_iters": 320000,
    "distributed_backend": "nccl",
    "lr_decay_style": "cosine",
    "warmup": 0.01,
--- a/megatron/model/mamba/mamba.py
+++ b/megatron/model/mamba/mamba.py
@@ -50,7 +50,10 @@ class ParallelMambaBlock(nn.Module):
         self.d_state = 16  # state dimensions per channel
         self.d_conv = 4  # convolution width
         self.expand = 2  # linear projection expansion factors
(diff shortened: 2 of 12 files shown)


Quentin-Anthony:
This looks good. Need to make it consistent with mamba and RWKV. We also need some TODO statements about revisiting this once we add swiglu. 

@jahatef is on it

jahatef:
Added support for mamba and RWKV, and added TODOs.

StellaAthena:
With these changes, is there a point to having separate Linear and LLaMA Linear definitions? At a glance it looks like all the differences are configurable, with the only difference being what is assumed as the default behavior if unspecified.

jahatef:
Refactored the activations and MLP layer to get rid of our redundant llama mlp class, and added some activations functions from https://arxiv.org/pdf/2002.05202.

configs/1-3B.yml
```diff
@@ -3,7 +3,7 @@
    # parallelism settings ( you will want to change these based on your cluster setup, ideally scheduling pipeline stages
    # across the node boundaries )
    "pipe_parallel_size": 1,
-   "model_parallel_size": 1,
+   "model_parallel_size": 2,
```
Quentin-Anthony, line 6: Let's not change the existing configs

Quentin-Anthony: The llama config changes are fine, to be clear

megatron/model/mamba/mamba.py
```diff
@@ -45,12 +45,18 @@ def __init__(
...
         ), "Mamba fused inner fn and bias in x_proj not compatible!"
 
+        assert neox_args.intermediate_size == None or neox_args.expansion_factor == None, "Must pass either the absolute intermediate size or the relative expansion factor for the mamba projections"
+
         # set variables, mostly following mamba defaults
         self.d_model = neox_args.hidden_size
         self.d_state = 16  # state dimensions per channel
         self.d_conv = 4  # convolution width
         self.expand = 2  # linear projection expansion factors
-        self.d_inner = int(self.expand * self.d_model)
```
Quentin-Anthony, line 53: Removing this will cause a failure at https://github.com/EleutherAI/gpt-neox/pull/1212/files#diff-cf396efcb6001846b18513bfb48e1e2681f3d240e589bfef9ea17cbdfe6b1218R80

megatron/model/mamba/mamba.py
```diff
@@ -45,12 +45,18 @@ def __init__(
...
+        if neox_args.intermediate_size == None:
+            neox_args.d_inner = neox_args.intermediate_size
+        if neox_args.expansion_factor:
+            self.expand = neox_args.expansion_factor
+        neox_args.d_inner = neox_args.intermediate_size
```
Quentin-Anthony, line 59: `neox_args.d_inner` is set both in the if-statement above, and also here?

megatron/model/rwkv/v6/rwkv.py
```diff
@@ -247,11 +247,11 @@ def __init__(self, neox_args, layer_number):
             self.time_maa_k = nn.Parameter(1.0 - torch.pow(ddd, ratio_1_to_almost0))
             self.time_maa_r = nn.Parameter(1.0 - torch.pow(ddd, ratio_1_to_almost0))
 
-        self.key = nn.Linear(neox_args.hidden_size, neox_args.dim_ffn, bias=False)
```
Quentin-Anthony, line 250: I prefer `ffn` to `ff`. Also, why are we swapping the ordering of `dim` and `ffn` in the first place?

megatron/model/transformer.py
```diff
@@ -150,84 +165,23 @@ def forward(self, hidden_states):
...
-        self.multiple_of = multiple_of
-
-        # Allow custom intermediate size, e.g. for Mistral
-        if neox_args.intermediate_size is not None:
-            ff_dim = neox_args.intermediate_size
-        else:
-            ff_dim = int(2 * neox_args.hidden_size * 4 / 3)
-            ff_dim = self.multiple_of * ((ff_dim + multiple_of - 1) // multiple_of)
-
-        self.w1 = mpu.ColumnParallelLinear(
```
Quentin-Anthony, line 192: need to test that tensor parallelism and MoE are maintained

megatron/model/transformer.py
```diff
@@ -1279,11 +1222,7 @@ def forward(self, x, attention_mask, layer_past=None):
                     raise KeyError(self.moe_type)
 
             with torch.enable_grad():
-                if (
-                    self.mlp_type == "llama"
-                    or self.num_experts > 1
-                    and self.moe_type == "deepspeed"
-                ):
+                if self.num_experts > 1 and self.moe_type == "deepspeed":
```
Quentin-Anthony, line 1225: shouldn't we still be testing if this mlp is a swiglu?

configs/1-3B.yml
jahatef: accidentally left it there when testing.

megatron/model/transformer.py
jahatef: I tested with TP, can run with MoE as well.

jahatef: I have tested MoE

Quentin-Anthony (approved):

PR #1214, opened 2024-05-10
fix pipeline parallelism detection

dmahan93:
All pythia configs come with pipe-parallel-size of 1 (e.g. [pythia-6.9b](https://github.com/EleutherAI/gpt-neox/blob/main/configs/pythia/6-9B.yml#L2)) and is also [set when it's greater than 1](https://github.com/EleutherAI/gpt-neox/blob/main/megatron/neox_arguments/arguments.py#L1036-L1039)

This fixes the detection to be in line with the is_pipe_parallel argument.


Quentin-Anthony:
Handled in https://github.com/EleutherAI/gpt-neox/pull/1218

PR #1218, opened 2024-05-16
Conversion script bugfixes

haileyschoelkopf:
Updates the NeoX-to-HF conversion utilities to fix the following problems:

- #1129 tweaks the default `is_pipe_parallel` behavior s.t. PP=1 models no longer are trained using PipelineModules, since MoE is not PP-compatible. This affects the save format of such checkpoints, so bringing the conversion script is_pipe_parallel detection logic in line with this.
- I found a bug where we would get `KeyError: no key 'final_linear.weight' could be found` when trying to convert a model with tied input/output embeddings. So adjusted the script to now save input embeds into the LM head state when converting a tied embedding model.


StellaAthena:
I think it would be more desirable to preserve the PP=1 -> PipelineModule with one pipeline stage behavior and have pinged the PR creator on Discord to discuss.

dmahan93:
should #1214 get closed and pulled into this?

haileyschoelkopf:
@dmahan93 yep!

I agree with Stella that going back to the behavior of pp=1 operating as a PipelineModule is desirable though.

Quentin-Anthony (approved):

PR #1221, opened 2024-05-16
Rwkv pipeline parallelism

jahatef:
Allow pipeline parallelism for rwkv.


Quentin-Anthony:
Simplest review of my life

Quentin-Anthony (approved):

PR #1226, opened 2024-05-21
Add Torch Profiler Support

DayOfThePenguin:
This turned out to be simpler to use and with the right arguments provides much of the same information as what's provided by nsight systems.  Bonus points for easy TensorBoard integration.

This might have to be rethought a bit to make it play nicely with WanB logging as well; I'm not familiar with how logging works there/if they'd support displaying these profiles natively.


Quentin-Anthony (dismissed):
Awesome. Tested and working for me.

Quentin-Anthony (approved):

PR #1240, opened 2024-06-21
SFT improvements (labeling fixes, different packing implementations)

dmahan93:
- add different packing impl (Unpacked, packing until overflow), a bit naive but for SFT it shouldn't be an issue
- fix labels to also have valid/test implementations
- fix label masking in _get_batch to also include anything from get_ltor_masks_and_position_ids


Quentin-Anthony (approved):
I've decided I'm happy with it as-is. Tested and working. Great work!

PR #1241, opened 2024-06-24
Fix paper reference in init_functions.py

rasbt:
I think this is supposed to be referencing "Transformers without Tears: Improving the Normalization of Self-Attention" ( https://arxiv.org/abs/1910.05895) which was published in 2019 not 2010


Quentin-Anthony:
Great catch. Thank you!

Quentin-Anthony (approved):

PR #1239, opened 2024-06-21
Add a chat data preprocessing script

dmahan93:
Add a chat based preprocessing script, to support multiturn conversation finetuning

PR #1242, opened 2024-06-25
Add DPO training

dmahan93:
Still a bit of a WIP (docs, adding precomputation of reference logprobs) but figured it'd be good to get it up here now since it's a fairly big change for any discussions that are needed.


dmahan93:
Also builds off https://github.com/EleutherAI/gpt-neox/pull/1240 and https://github.com/EleutherAI/gpt-neox/pull/1239 since the packing implementations/chat templating items are much more useful for DPO.

Quentin-Anthony (approved):
Passes my tests. Great work @dmahan93 !

PR #1247, opened 2024-06-28
Add hf llama to neox conversion

dmahan93:


dmahan93:
FYI Does not support pipeline parallelism

Quentin-Anthony (approved):
Tested and working for me. Had to fix some pre-commit formatting that must have snuck in from a previous PR.

PR #1257, opened 2024-08-08
Megatron-LM style Sequence Parallel

### LINKED ISSUE #812: Add support for sequence parallelism
Add sequence parallelism from upstream megatron

- https://github.com/search?q=repo%3ANVIDIA%2FMegatron-LM+sequence_parallel&type=code
- https://arxiv.org/pdf/2205.05198.pdf

haileyschoelkopf:
closes #812 .

This PR aims to add support for Reduce-Scatter type Tensor Parallelism (aka sharding LNs across the TP group), as used in Megatron-LM and as described by https://arxiv.org/abs/2205.05198 . 

Current progress:
- [x] Modify ColumnParallelLinear and RowParallelLinear to handle the necessary comms for SP
- [x] [WIP] Keep LN gradients synchronized across TP ranks, by allreducing their grads across the TP group as well as the DP group
- [x] Test convergence, throughput, and memory usage against non-SP tensor parallel baseline
- [x] Other compatibility checks, such as compatibility with PP, MoE, ...

Commit dc4c99be59fc49029f1f18fd91200bae20e91bee allows one to run train.py when sequence_parallel is enabled, but convergence is very poor because the grads of the LNs are not synchronized over TP ranks. (loss falls slower, and ends up flatlining around ~8 after 100 or so updates.)

Commit 3ccd3ba9faaf02ec7f6760823403e1ca436687f8 does 2 things: 
1) adds a function that adds backward hooks to the LN parameters, which are *supposed* to be reducing gradients across the TP region, but these hooks do not seem to ever trigger (TBD why this is happening--the hooks definitely do get applied.)
2) sanity-checks the SP comms added in `megatron.mpu.mappings` work, by implementing regular TP via having the RowParallelLinear do Reduce-Scatter immediately followed by All-Gather. This does indeed give the same results as regular TP, so I feel fairly confident that the core SP comms and Row/ColumnParallel logic was implemented correctly.

commit 92ed0ccfc1ded09db2d9606d76f6809c6c8da010 (most recent commit atm) does do Sequence Parallelism and still shows the same convergence issues, despite the hooks on the LNs that should theoretically sync their grads. These hooks do get added when we run the `megatron.utils.mark_norms...` function when I checked this, but they don't ever seem to get run. Hooks that DeepSpeed itself adds in a similar manner do trigger when ZeRO stage 2 is run.

I'll also push a commit soon in which I tried to add `mpu.get_sequence_parallel_group()` and `mpu.get_sequence_data_parallel_group()` functions and distributed groups upon MPU initialization--DeepSpeed ZeRO optimizers use the Sequence+Data parallel group for sharding and grad allreduce when the MPU exposes this: https://github.com/microsoft/DeepSpeed/blob/324ee65cb0e5592cfa3a4d82273b2cd952b10a93/deepspeed/runtime/engine.py#L1138 however, it doesn't seem like this actually fixed any behavior when I implemented it. It's also used for DeepSpeed-Ulysses, so I don't think this is the way to go regardless as it won't likely do precisely what we want, though I wanted to rule out it fixing the convergence issues I was seeing.

WandB runs testing this branch with various edits can be found here: [https://wandb.ai/schoelkopf/neox-sequence-parallel/workspace](https://wandb.ai/schoelkopf/neox-sequence-parallel/workspace).


haileyschoelkopf:
@bclyang has fixed convergence with #1260 #1259 ! See #1260 for convergence tests run.

It's no longer useful for this PR, but for posterity I've pushed to branch `seq-dp-mpu-group` the commit where I tried to add the `mpu.get_sequence_parallel_group(), mpu.get_sequence_data_parallel_group()` functions to the MPU, even though we aren't intending to add DS-Ulysses.

megatron/model/transformer.py
```diff
@@ -113,7 +113,7 @@ def __init__(
             neox_args=neox_args,
             input_size=neox_args.hidden_size,
             output_size=ff_dim,
-            gather_output=False,
+            gather_output=False,  # TODO: add a parallel-input check? need to AG in fwd pass to reshard, here?
```
Quentin-Anthony, line 116: Can you describe further? Is this still needed?

configs/neox_arguments.md
```diff
@@ -1056,6 +1056,16 @@ Parallelism Arguments
 
 
 
+- **sequence_parallel**: bool
```
haileyschoelkopf, line 1059: Is this what we want the arg to be called? 

do we want there to be an `is_sequence_parallel` boolean flag that gets set, a la `is_pipe_parallel`? (I think it's not necessarily needed but happy to do so)

configs/test-seqpar/410M_baseline.yml
```diff
@@ -0,0 +1,106 @@
+{
```
haileyschoelkopf, line 1: TODO before merge: delete this test config.

megatron/model/utils.py
```diff
@@ -359,3 +352,45 @@ def get_fusion_type(neox_args):
...
+    register a hook on each parameter which will allreduce norms' weights' grads across
+    the model (sequence) parallel region.
+    """
+
+    if not neox_args.sequence_parallel:
+        # if we aren't using sequence parallelism, this is a no-op
+        return
+
+    for module_ in module.modules():
+        if "norm" in type(module_).__name__.lower():
```
haileyschoelkopf, line 392: We can alternately commit to applying this hook to any parameter for which `param.sequence_parallel` is marked as True. 

This might be more future proof.

cc @Quentin-Anthony

haileyschoelkopf: Ah sorry, actually the reason I did not do this is because for e.g. the built-in Torch LayerNorm, it'd be slightly annoying to mark its parameters on the fly.

bclyang:
Here are the WandB runs for the convergence tests:

Baseline ([`410M_baseline](https://github.com/EleutherAI/gpt-neox/pull/1257/files#diff-2e65bd64b6c7c3953e544937d448ce797689c68caa972d122679fc66957c5eb9)
- [sequence_parallel=False]https://wandb.ai/brandony/neox/runs/e3ou7n5v?nw=nwuserbrandony
- [sequence_parallel=True]https://wandb.ai/brandony/neox/runs/ywh9p3ks?nw=nwuserbrandony

Weight tying (`no_weight_tying=False`)
- [sequence_parallel=False]https://wandb.ai/brandony/neox/runs/yfe3dlrj?nw=nwuserbrandony
- [sequence_parallel=True]https://wandb.ai/brandony/neox/runs/a1rp6les?nw=nwuserbrandony

Pipeline Parallel `pipe_parallel_size=1`
- [sequence_parallel=False]https://wandb.ai/brandony/neox/runs/ja4rlqhr?nw=nwuserbrandony
- [sequence_parallel=True]https://wandb.ai/brandony/neox/runs/0gg5sfyx?nw=nwuserbrandony

megatron/model/utils.py
```diff
@@ -359,3 +353,45 @@ def get_fusion_type(neox_args):
...
+    Allreduces grads for e.g. LN weights across the model parallel group.
+    Needed to keep LNs in sync, despite them getting diff data -> diff gradients when using sequence parallel.
+    """
+    # Bypass the function if no TP -> no comm needed.
+    if mpu.get_model_parallel_world_size() == 1:
+        return input_
+
+    # Bf16 convert
+    dt = input_.dtype
+    if dt == torch.bfloat16 and mpu.get_fp32_allreduce():
```
Quentin-Anthony, line 369: what about fp16?

megatron/model/utils.py
Quentin-Anthony: Yeah I'm fine with sticking with this as-is.

megatron/mpu/mappings.py
```diff
@@ -98,6 +98,82 @@ def _gather(input_):
...
+def _reduce_scatter_along_seq_dim(input_, seq_dim):
+    """Reduce-scatter the input tensor across model parallel group, scattering across sequence dim."""
+    world_size = get_model_parallel_world_size()
+    # Bypass the function if we are using only 1 GPU.
+    if world_size == 1:
+        return input_
+
+    # Bf16 convert
+    dt = input_.dtype
+    if dt == torch.bfloat16 and get_fp32_allreduce():
```
Quentin-Anthony, line 110: what about fp16?

megatron/mpu/mappings.py
```diff
@@ -98,6 +98,82 @@ def _gather(input_):
...
+    """Gather tensors and concatinate along the (manually-specified) sequence dimension."""
+
+    world_size = get_model_parallel_world_size()
+    # Bypass the function if we are using only 1 GPU.
+    if world_size == 1:
+        return input_
+
+    # Bf16 convert
+    dt = input_.dtype
+    if dt == torch.bfloat16 and get_fp32_allreduce():
```
Quentin-Anthony, line 139: fp16?

configs/neox_arguments.md
Quentin-Anthony: I like this as-is. No need for an `is_sequence_parallel` unless later on we need to handle such models differently like in the pipeline case.

megatron/training.py
```diff
@@ -835,6 +838,7 @@ def backward_step(neox_args, timers, optimizer, model, loss):
 
 def train_step(neox_args, timers, data_iterator, model, optimizer, lr_scheduler):
     """Single training step."""
+    modules_dict = dict(model.named_modules())
```
Quentin-Anthony, line 841: what's this for?

megatron/mpu/mappings.py
haileyschoelkopf: good point!

This actually impacts our other existing comms pre-PR (see e.g. `_reduce`), shall we fix them all in this PR? could add that change in a separate one if preferred.

megatron/training.py
haileyschoelkopf: oops, missed this in cleanup. @bclyang this was just for debugging via printing parameter values right?

megatron/mpu/mappings.py
haileyschoelkopf: Made the change for all functions! commit (https://github.com/EleutherAI/gpt-neox/pull/1257/commits/ab11a6a71c122cc67440b2206ebe81cfaa70d242) can be cherrypicked into a new PR if preferred.

megatron/mpu/mappings.py
haileyschoelkopf: See https://github.com/EleutherAI/gpt-neox/pull/1257#discussion_r1722505300

megatron/model/utils.py
haileyschoelkopf: see https://github.com/EleutherAI/gpt-neox/pull/1257#discussion_r1722505300

megatron/training.py
bclyang: yes, thanks!

megatron/mpu/mappings.py
Quentin-Anthony: We can keep it in this one

bclyang:
Tested saving and loading checkpoints with sequence parallel, see this WandB of a resumed run: https://wandb.ai/brandony/neox/runs/4tac7i8k?nw=nwuserbrandony.

Quentin-Anthony (approved):
Reviewed, tested, and working!

PR #1259, opened 2024-08-13
Fix LayerNorm all reduce gradient hook

bclyang:
Fix the all reduce gradient hook for LayerNorm, by moving hook registration for gradient all reduce for the LayerNorm parameters to after `deepspeed.initialize`. Update the gradient hook to sum in the LayerNorm gradient in the sequence dimension instead of average.

Test Plan
- [x]  Printed gradients and verified they're now synced across GPUs
- [x]  Printed LayerNorm parameters after 10 steps and verified they're now in sync

This PR does not fix the convergence issue, since the loss still ends up flat-lining around 7.5 for enwiki8. When printing out LayerNorm parameters, I observed that in the current `sequence_parallel=True` implementation, the LayerNorm weight parameters change much slower than when `sequence_parallel=False`.

### DIFF
--- a/configs/neox_arguments.md
+++ b/configs/neox_arguments.md
@@ -111,7 +111,7 @@ Logging Arguments
 
 - **git_hash**: str
 
-    Default = c93c1b4
+    Default = 9a43318
 
     current git hash of repository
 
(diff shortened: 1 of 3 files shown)


haileyschoelkopf (approved):
fantastic work! merging into the working PR, though convergence is still unsolved.

PR #1262, opened 2024-08-20
mamba fixes and cleaning

jahatef:
adds:
- some common configs to mamba configs so they can be ran easier.
- a fix to some mamba configs
- bypasses some checks in arguments.py
- an assertion error when mamba is missing packages instead of a warning message

### DIFF
--- a/megatron/model/mamba/mamba.py
+++ b/megatron/model/mamba/mamba.py
@@ -3,7 +3,6 @@ import math
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-
 try:
     from mamba_ssm.ops.selective_scan_interface import (
         selective_scan_ref,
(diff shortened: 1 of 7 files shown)


megatron/model/mamba/mamba.py
```diff
@@ -13,10 +13,8 @@
     from causal_conv1d import causal_conv1d_fn
     import einops
 except ModuleNotFoundError:
-    print(
-        "Unable to import Mamba kernels. Install them from our requirements/requirements-mamba.txt, or directly from https://github.com/state-spaces/mamba"
-    )
-    pass
+    assert False, "Unable to import Mamba kernels. Install them from our requirements/requirements-mamba.txt, \
```
haileyschoelkopf, line 16: This will assert for all users, not just those wanting to run Mamba models. We should move this to only trip the assert if a Mamba layer is initialized and we still get a ModuleNotFoundError

jahatef: I just removed this change for now, we can add something later but I think the other changes are more important

jahatef:
removed the assertion message for now.

Quentin-Anthony (approved):
LGTM

PR #1265, opened 2024-08-29
add Apex fused RMS norm

dmahan93:
Adds Apex's fused RMS norm


Quentin-Anthony:
merged as part of https://github.com/EleutherAI/gpt-neox/pull/1269

PR #1269, opened 2024-09-06
LayerNorm Refactor

aurelion-source:
Adds:
- Apex's rmsnorm
- TransformerEngine's rmsnorm and layernorm

PR #1272, opened 2024-09-09
TE Import Hotfix

Quentin-Anthony:
Change the TE import to be properly conditional. I didn't catch it in https://github.com/EleutherAI/gpt-neox/pull/1269 since I had tested it with TE installed (oops!)

Refactored this function while I was around.

@aurelion-source -- FYI

### DIFF
--- a/megatron/model/utils.py
+++ b/megatron/model/utils.py
@@ -25,8 +25,15 @@ from types import GeneratorType
 import torch.distributed as dist
 
 
-def get_params_for_weight_decay_optimization(module, neox_args):
-    """Divide params into with-weight-decay and without-weight-decay groups.
+import importlib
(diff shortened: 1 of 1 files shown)


aurelion-source:
> Change the TE import to be properly conditional. I didn't catch it in #1269 since I had tested it with TE installed (oops!)
> 
> Refactored this function while I was around.
> 
> @aurelion-source -- FYI

@Quentin-Anthony 
#1269 was tested with and w/o TE and Apex. It looks like this line was overwritten by a commit after my rebase onto main https://github.com/aurelion-source/gpt-neox/blob/d1dc7d7228d93b684c55d4bea0885829e010a9e2/megatron/model/utils.py#L41

PR #1274, opened 2024-09-09
Fix weight decay module check

aurelion-source:
- Updates is_no_weight_decay_module to check for module name.
- Adds checks for MixedFused (Apex) layers.
- Remove redundant TE imports.


Quentin-Anthony (approved):
Thanks for catching this!

Issue #1268, opened 2024-09-06
Allow training without knowing num_iters


StellaAthena:
**Is your feature request related to a problem? Please describe.**
I think there's a misalignment between library design and how people typically think about training models. People typically think about how they have a fixed dataset and want to train a model on it, not that they have a bunch of data and have a fixed training **length** they would like to run for. This can cause problems with the fact that we force people to manually calculate the number of training iters.

**Describe the solution you'd like**
Allow setting train_iters automatically, perhaps by allowing an "auto" argument and then using the formula

train_iters * seq_length * mbs * grad_accumulation * data_parallel_size = num_tokens

internally. Additionally, when using "auto" we may wish to allow a "num_epochs" argument. We should make sure that the behavior when setting "num_epochs" is to do the entire dataset once and then the entire dataset again, not replicating the data and then shuffling all the sequences together.

We could also add warnings for mistakes that are easier to make when num_iters is implicitly set, such as when the warm_up is greater than 0.1*num_iters (highly unlikely to be desired) or when its greater than num_iters (which is wrong)

**Describe alternatives you've considered**
An auxillery function that computes num_iters for you would have all the same benefits but the additional downside of adding an extra step for users.

**Additional context**
Aflah and SE Gyes have both expressed confusion about this in the discord sever, as have users at at least one national lab.

Quentin-Anthony:
Agreed. Assigned to @aurelion-source.

PR #1283, opened 2024-09-16
Automatically compute train_iters when train_epochs is specified.

AI-WAIFU:
Major changes:

 - Introduces a train_epochs argument
 - Reorders events in the pretrain function, changing from: model/optimizer creation -> data iterator creation to: data loader creation -> model/optimizer creation -> data iterator creation
 - Removes save_iters property from NeoX args and replaces it with a runtime is_save_iter function
 - Adds support for non integer checkpoint factors when using logarithmic checkpointing
 
 
 
Note: Most of these changes are a consequence of not being able to compute train_iters when creating the NeoX Args object. At a high level we pass both train_epochs and train_iters down to the dataloader, and use the one that is not none to specify the dataloader behavior, then if train_iters is unspecified we infer it from the dataloader after constructing it.


Quentin-Anthony:
fixes https://github.com/EleutherAI/gpt-neox/issues/1268

Quentin-Anthony (approved):
Tested and working for me.

PR #1292, opened 2024-09-23
update neox arg docs

Quentin-Anthony:
I believe we're out of github runner credits so this isn't automatic anymore.

FYI -- @jahatef


Quentin-Anthony:
oops was on a fork. Redoing this from main.

PR #1296, opened 2024-09-29
hotfix for tp >= 2 and pp > 2 in autoitercount

AI-WAIFU:
Add's broadcasting of train_iters from rank 0 when dataloader is not present


Quentin-Anthony (approved):
Working for me. Merging.

PR #1282, opened 2024-09-16
TransformerEngine Integration

aurelion-source:
Adds:

- TELinear
- TELayerNormMLP
- TEColumnParallelLinear
- TERowParallelLinear
- TEMultiheadAttention
- TEDelayedScaling (in progress)

PR #1297, opened 2024-09-30
TE integration via full TransformerLayer

tf-nv:
This is a sketch of using the attention picking mechanism ("global", "flash", NEW: "TE") to use the high level TransformerLayer from TransformerEngine. This is more of a prototype to show that integration with deepspeed is possible and what perf to expect.

Things that work:
1. TE attention TFLOPS are 5% higher than flash attention in BF16, and 70% higher in FP8
2. Training an 22B model on multiple DGXH100 with zero 1 and TP2 (BF16)
3. Activation checkpointing from TE

Many aspects are hardcoded, e.g. RoPE and activation checkpointing can not be reconfigured from the config files. #1282 is much more elaborate in that it exposes TE layers on a much lower level. Meanwhile this PR could serve as a benchmark, showing what is possible with TE on a classic GPT2 style network.

I kept the implementation as minimal as possible, there is room for further performance depending on the workload. There is e.g. sequence parallelism and different memory layouts.

The dockerfile now uses a later ngc pytorch container and installs a later deepspeed tag from source for compatibility.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
 # Copyright (c) 2024, EleutherAI
+# This file is based on code by the authors denoted below and has been modified from its original version.
+#
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM nvcr.io/nvidia/pytorch:24.02-py3
+FROM nvcr.io/nvidia/pytorch:24.06-py3
 
 ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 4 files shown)


Quentin-Anthony:
Will merge this after the finegrained TE PR

megatron/model/transformer.py
```diff
@@ -1301,6 +1309,38 @@ def forward(self, args):
...
+        )
+        PE = RotaryPositionEmbedding(dim=hidden_size_per_attention_head)
+        self.rotary_pos_emb = PE(kwargs["seq_length"]).to(device="cuda")
+
+    def forward(self, args):
+        assert len(args) == 2, ("TE transformer layer pipe must have two args, hidden_states and the attention mask. "
+                                "The mask will be discarded")
+        hidden_states, attention_mask = args
+
+        fp8_format = Format.HYBRID  # E4M3 during forward pass, E5M2 during backward pass
```
Quentin-Anthony, line 1337: should instead accept neox_arg from the new `te_fp8_format`

megatron/model/gpt2_model.py
```diff
@@ -271,6 +272,24 @@ def init_specs(self):
                         layer_number=i,
                     )
                 )
+            elif layer_type in ["TE"]:
```
Quentin-Anthony, line 275: needs tested with PP and TP, since we'd be relying on two external codebases (deepspeed for PP, TE for TP) whose topologies probably don't play nicely together.

Dockerfile
```diff
@@ -82,6 +85,12 @@ COPY megatron/fused_kernels/ /megatron/fused_kernels
 WORKDIR /megatron/fused_kernels
 RUN python setup.py install
 
+SHELL ["/bin/bash", "-c"]
+
+RUN DS_BUILD_FUSED_LAMB=1 DS_BUILD_FUSED_ADAM=1 DS_BUILD_TRANSFORMER=1 DS_BUILD_STOCHASTIC_TRANSFORMER=1  DS_BUILD_UTILS=1 \
+    TORCH_CUDA_ARCH_LIST="8.0 9.0+PTX" \
+    python -m pip install git+https://github.com/microsoft/DeepSpeed.git@v0.14.4
```
Quentin-Anthony, line 92: this should probably instead be latest [deeperspeed](https://github.com/EleutherAI/DeeperSpeed). Can't hardcode the arch list.

PR #1308, opened 2024-10-17
Add ERROR logging prefix and sort the prefixes alphabetically

TheBatmanofButler:
When trying to run `./deepy.py train.py ...` I got the following error:

> `NameError: name 'ERROR' is not defined`

I tracked this down and found that we have a number of formatted logging prefixes with colors, but `ERROR` was missing. I added that (with the color red, same as for `FAIL`), sorted the prefixes alphabetically, and grouped them a bit for clearer organization. Tested and got an actual error message after the change.

### DIFF
--- a/megatron/neox_arguments/arguments.py
+++ b/megatron/neox_arguments/arguments.py
@@ -50,16 +50,19 @@ from .neox_args import (
     ATTENTION_TYPE_CHOICES,
 )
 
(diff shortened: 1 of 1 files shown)


Quentin-Anthony:
Thanks for the catch!

Quentin-Anthony (approved):

Issue #1314, opened 2024-10-28
Training crashes when "(hidden_size * num_kv_heads) / (num_attention_heads * num_attention_heads)" is not an integer.


tiandeyu-cs:
**Describe the bug**
Training crashes when "**(hidden_size * num_kv_heads) / (num_attention_heads * num_attention_heads)**" is not an integer.

**To Reproduce**
Train a model with configuration as follows:

```
{
	"hidden_size": 5120,
	"num_attention_heads": 40,
	"num_kv_heads": 8,
	"seq_length": 4096,
	"batch_size": 1
}
```

It will crash with the following error:

> RuntimeError: shape '[4096, 1, 5, 179]' is invalid for input of size 3670016

PR #1315, opened 2024-10-28
fix a GQA issue (#1314)

tiandeyu-cs:
fix a GQA issue (#1314)
- do not create a fake head dim and split the 'mixed_x_layer' into QKV layers directly.

### DIFF
--- a/megatron/model/transformer.py
+++ b/megatron/model/transformer.py
@@ -763,51 +763,16 @@ class ParallelSelfAttention(nn.Module):
         # pass through projection: [sq, b, h] --> [sq, b, ((np + 2 * kvp) * hn)]
         mixed_x_layer, _ = self.query_key_value(hidden_states)
 
-        # First: reshape so we have seqlen, batch, and num. query heads each as separate dims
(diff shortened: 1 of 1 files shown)


Quentin-Anthony (approved):
Working for me. Nice catch.

aflah02:
Hi @tiandeyu-cs 
I am facing issues while exporting a GQA model to HF (it exports but produces gibberish while the non-GQA works perfectly). I was wondering if this PR has any implications on the export scripts if you've trained models using NeoX with GQA

tiandeyu-cs:
This PR does not change GQA models; it only fix a bug in __calculating__ GQA. I do not think this PR can affect model exporting.

aflah02:
> This PR does not change GQA models; it only fix a bug in **calculating** GQA. I do not think this PR can affect model exporting.

Thanks! I actually used your PR as reference and asked ChatGPT to make the corresponding changes in the export file. It cooked up some incorrect versions but the final version now converts my GQA models properly and they output coherent stuff while earlier it generated gibberish post conversion. I honestly was too sleepy and didn't look too closely apart from seeing inference results but yeah

PR #1316, opened 2024-10-30
Update text_generation_utils.py to work with pipe_parallel_size of 0

markNZed:
During training on a single GPU to get the evaluation step running we set "pipe_parallel_size": 0 but this created a problem in text_generation_utils.py  that was not handling the data structure returned when the parallel pipeline is disabled.


aurelion-source:
Resolved in #1336

PR #1332, opened 2025-01-04
Adding Support for Non-Parametric Layernorm

aflah02:
Hi
Based on our discord discussion I've added non-parametric layernorm as described in the OLMo paper.


megatron/model/norms.py
```diff
@@ -33,6 +33,12 @@ def get_norm(neox_args):
             norm = MixedFusedLayerNorm
         else:
             norm = LayerNorm
+    elif neox_args.norm == "non_parametric_layernorm":
+        eps = neox_args.layernorm_epsilon
+        if neox_args.layernorm_fusion:
+            raise ValueError(f"neox_args.layernorm_fusion not supported for non_parametric_layernorm")
+        else:
+            norm = LayerNorm(elementwise_affine=False, bias=False)
```
aurelion-source, line 41: This leads to an error due to initializing torch.nn.LayerNorm without normalized_shape.
Also, get_norm should not return an initialized object.
#1338 resolves this by creating a separate NonParametricLayernorm class.

PR #1341, opened 2025-02-28
Update inspect_checkpoints.py

aflah02:
Fixed typo which raised errors when supplying dir

### DIFF
--- a/tools/ckpts/inspect_checkpoints.py
+++ b/tools/ckpts/inspect_checkpoints.py
@@ -295,7 +295,7 @@ def compare(args: Namespace):
 def main():
     parser = ArgumentParser()
     parser.add_argument(
-        "dir",
+        "--dir",
         type=str,
         help="The checkpoint dir to inspect. Must be either: \
          - a directory containing pickle binaries saved with 'torch.save' ending in .pt or .ckpt \


aflah02:
There are some some issues around this line - https://github.com/EleutherAI/gpt-neox/blob/17bed2a7b62d167aa096d820d84be38c65be66f5/tools/ckpts/inspect_checkpoints.py#L58

I am currently using this hacky fix - 
```
    try:
        col_size = max(len(str(k)) for k in contents)
    except Exception as e:
        print(f"Error: {e}")
        col_size = 100
```

aurelion-source:
> Fixed typo which raised errors when supplying dir

I don't think there's a typo. dir is supposed to be a positional argument as explained in the readme here: https://github.com/EleutherAI/gpt-neox/tree/main/tools/ckpts#inspect_checkpointspy

> There are some some issues around this line -
> 
> https://github.com/EleutherAI/gpt-neox/blob/17bed2a7b62d167aa096d820d84be38c65be66f5/tools/ckpts/inspect_checkpoints.py#L58
> 
> I am currently using this hacky fix -
> 
> ```
>     try:
>         col_size = max(len(str(k)) for k in contents)
>     except Exception as e:
>         print(f"Error: {e}")
>         col_size = 100
> ```

Thanks for flagging this! #1346 resolves it by skipping pretty_print calls when the dict is empty.

Issue #1337, opened 2025-02-14
convert_hf_llama_to_neox.py appears incompatible with LLAMA-3.1-70B


tijmen:
## Context and Issue Description

I am trying to fine tune Llama-3.1-70B on ORNL Frontier and I'm getting a loss of ~12 on a small sample of fineweb, which is obviously much too large. Suspecting incorrect loading of the base model into GPT-NeoX, I tried running some text generation where I find that the outputs of the model look like completely garbled. I suspect either a misconfiguration on my part, or an incompatibility of the convert_hf_llama_to_neox.py script with LLAMA-3.1-70B. 

## Details

I converted the base model into a tp==8, pp==0 copy of the model in the format expected by GPT-NeoX with
```bash
python $HOME/gpt-neox/tools/ckpts/convert_hf_llama_to_neox.py --model $HOME/models/llama-3.1-70b --model_path $HOME/models/llama-3.1-70b-neox --tp 8
```
I then ran unconditional text generation using the following config files

```yaml
{
  "vocab_file": "/ccs/home/tijmen/models/llama-3.1-70b/tokenizer.json",
  "tokenizer_type": "HFTokenizer", 
  "num_layers": 80,
  "hidden_size": 8192,
  "intermediate_size": 86016,
  "num_attention_heads": 64,
  "num_kv_heads": 8,
  "seq_length": 8192,
  "max_position_embeddings": 8192,
  "pos_emb": "rotary",
  "rotary_pct": 1,
  "rotary_emb_base": 500000,
  "no_weight_tying": true,
  "gpt_j_residual": false,
  "output_layer_parallelism": "column",
  "norm": "rmsnorm",
  "rms_norm_epsilon": 1.0e-5,
  "use_bias_in_norms": false,
  "use_bias_in_attn_linear": false,
  "use_bias_in_mlp": false,
  "activation": "swiglu",
  "mlp_multiple_of": 256,
  "make_vocab_size_divisible_by": 1,
  "bf16": {
    "enabled": true
  },
  "data_types": {
    "grad_accum_dtype": "fp32"
  },
  "attention_config": [[["flash"], 80]],
  "scaled_upper_triang_masked_softmax_fusion": true,
  "bias_gelu_fusion": false,
}
```

```yaml
{
  "launcher": "slurm",
  "deepspeed_slurm": true,
  "load": "/ccs/home/tijmen/models/llama-3.1-70b-neox",
  "no-load-optim": true,
[...]

PR #1342, opened 2025-03-03
RMSNorm epsilon implementation

tijmen:
Make RMSNorm implementation match the huggingface transformers one. epsilon now adds to the variance rather than adding to the RMS. This fixes compatibility with Llama but could break compatibility with any models using RMSNorm and trained with previous versions of GPT-NeoX. 

See #1337 for more info.


Quentin-Anthony:
addressed by https://github.com/EleutherAI/gpt-neox/pull/1345

PR #1356, opened 2025-04-26
fix bug in EvalHarnessAdapter when pipe parallel size is set to 1

### LINKED ISSUE #1320: _forward_step_fn does not always return two values so eval.py breaks if is_pipe_parallel is false
This call to _forward_step_fn expects two values returned  https://github.com/EleutherAI/gpt-neox/blob/59a5236ddaf721890e3d6ef98fb8ca66c2266ce0/eval_tasks/eval_adapter.py#L372

The forward_step can return three values https://github.com/EleutherAI/gpt-neox/blob/59a5236ddaf721890e3d6ef98fb8ca66c2266ce0/megatron/training.py#L847

I guess I am seeing this because I have is_pipe_parallel false and that is uncommon. Maybe there needs to be an option not to return metrics.

MarisaJH:
resolves https://github.com/EleutherAI/gpt-neox/issues/1320

### DIFF
--- a/eval_tasks/eval_adapter.py
+++ b/eval_tasks/eval_adapter.py
@@ -59,8 +59,8 @@ class EvalHarnessAdapter(HFLM):
         self.is_main = neox_args.rank == 0
         self.is_local_main = neox_args.local_rank == 0
         self.is_model_parallel = neox_args.model_parallel_size > 1
-        self.is_pipe_parallel = self.model.is_pipe_parallel
-        self.is_data_parallel = self.model.is_data_parallel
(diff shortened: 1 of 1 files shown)


aflah02:
I didn't realize there is a PR but came up with the same changes haha
Stumbled across this randomly today

StellaAthena:
@Quentin-Anthony can we get this reviewed and/or merged?

Quentin-Anthony:
Yes I'll review this today

Quentin-Anthony (approved):

PR #1357, opened 2025-05-12
Fix neoxargs consume without config dir test

emmanuel-ferdman:
# PR Summary
Currently `test_neoxargs_commandline.py` contains two tests with same name `test_neoxargs_consume_deepy_args_with_config_dir`. This PR fixes the neoxargs consume without config dir test so it won't be silently ignored.


Quentin-Anthony:
Thanks for catching this!

PR #1372, opened 2026-02-11
fix: undefined self.args in TopKTokenChoiceRouter.jitter

Mr-Neutr0n:
## Bug

`TopKTokenChoiceRouter.jitter()` references `self.args.moe_jitter_eps`, but `self.args` is never assigned in the constructor. This causes an `AttributeError` at runtime when jitter noise is applied during training.

The constructor already stores the value as `self.jitter_eps = neox_args.moe_jitter_eps`, but the `jitter()` method does not use it.

## Fix

Changed `self.args.moe_jitter_eps` to `self.jitter_eps` in the `jitter()` method to reference the attribute that is actually set in `__init__`.

### DIFF
--- a/megatron/model/router.py
+++ b/megatron/model/router.py
@@ -223,8 +223,8 @@ class TopKTokenChoiceRouter(torch.nn.Module):
         Returns:
             torch.Tensor: Jittered input tensor.
         """
-        low = 1.0 - self.args.moe_jitter_eps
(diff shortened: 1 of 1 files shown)


Quentin-Anthony:
this was included along with https://github.com/EleutherAI/gpt-neox/commit/ea7aefd8f8e183256992d9fede4d7c71a46d67a5

PR #1383, opened 2026-05-16
Add route to modify megatron fused kernel build path

aflah02:
Adding this as the default value oftens leads to collisions during multi-node usage.

I had hardcoded it to `/tmp/...` and having the ability to change it via the config is better. Default behaviour is preserved if the arg is unused


Quentin-Anthony (approved):

Quentin-Anthony:
Passes my testing

PR #1384, opened 2026-05-18
Fix TE Crash with RMSNorm

aflah02:
As shared on Discord


Quentin-Anthony (approved):
thanks for catching this!

PR #1388, opened 2026-06-02
Torch Run Launcher

aflah02:
Add support for torchrun based launching on SLURM machines


aflah02:
It might also be of interest to keep a dir called launchers where people can contribute their own custom launchers depending on their cluster's weird requirements. Not all launchers would support everything but maybe that's better than crowding the main repo? or should it go here like the JSRun example? https://github.com/EleutherAI/DeeperSpeed/commit/9aed6c8500d7c492d85c5c88687322dbda70e370

aflah02:
This will be paired with a DeeperSpeed PR and needs changes to migrate some of this logic to DeeperSpeed
Moving to draft until that is ready

PR #1389, opened 2026-06-11
Align Args in .md with .py files

aflah02:
As discussed on discord 

## Summary

  Adds a static NeoX argument docs sync utility and PR check to keep configs/neox_arguments.md aligned with the Python argument source files.

  ## Changes

  - Added configs/neox_args_sync.py to check or regenerate configs/neox_arguments.md from:
      - megatron/neox_arguments/neox_args.py
      - megatron/neox_arguments/deepspeed_args.py

  - Added --sync mode:

    python configs/neox_args_sync.py --sync

  - Added a GitHub Action that runs on PRs and blocks if args are not in sync

  - Updated configs/neox_arguments.md to match the Python sources.
  - Added contributor documentation explaining how to sync argument docs.
  - Replaced the old import-based docs workflow with a static parser approach that does not require importing NeoX or installing torch.
  - git_hash is still ignored for sync validation, but --sync renders its current git describe --always value.

## Open Questions -

Should I just remove gen_docs.py?

Do we need git_hash?


Quentin-Anthony:
I fixed some edge cases:
- detect arguments documented under the wrong NeoXArgs section
- reject duplicate markdown argument entries instead of silently overwriting them
- prevent --neox-only --sync from deleting the DeepSpeed sections
- support Python 3.8 in addition to 3.9+ by avoiding removeprefix() and ast.unparse()
- render empty-string defaults explicitly as ''
- render git_hash as <dynamic> instead of a stale commit hash
- document the new git_hash sync behavior. Looks like this was broken before.

Quentin-Anthony:
rebased on upstream just now, hence the repeated commits and force push you may see

Quentin-Anthony (approved):

PR #1392, opened 2026-06-17
Add ability to log loss/PPL per validation/test set

aflah02:
As discussed on Discord

The PR would benefit from first merging https://github.com/EleutherAI/gpt-neox/pull/1387, https://github.com/EleutherAI/gpt-neox/pull/1389 and https://github.com/EleutherAI/gpt-neox/pull/1356

These PRs fix eval which is currently fully broken and all runs crash when pipe parallel size is set to 1. One of the PRs also aligns args which is not contingent but would be nice to align args once and for all for future PRs.

In my testing it works well and logs separately as expected. Design Doc is just for reference and will be removed post review/ can be removed even before that if preferred

### DIFF
--- a/configs/neox_arguments.md
+++ b/configs/neox_arguments.md
@@ -1227,6 +1227,20 @@ Training Arguments
 
     List of paths to validation reward datasets
 
+- **valid_data_names**: list
+
+    Default = None
+
+    Optional names for validation dataset subsets when eval_loss_logging uses separate logging.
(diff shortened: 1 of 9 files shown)


aflah02:
Tests pass locally :)

aflah02:
Some changes in the last commit also fix other tests locally and are unrelated to the core PR but were existing bugs - Happy to move to another PR if that is preferred

Quentin-Anthony (approved):

Issue #1394, opened 2026-06-18
Tracker: Improve HF Integration


aflah02:
Issue to track support for reference config files and scripts to load in and export models from the following series (both with and without Transformer Engine). For now, I'll focus only on the dense offerings from each generation and only those models which can be constructed with existing components and don't require separate PRs to bring in.

Models Under Consideration:

# Model Support Tracker

##  Qwen

| Series   | Config | Import Script | Export Script | Tests | Works with Transformer Engine | Remarks |
|----------|--------|---------------|----------------|-------|-------------------------------|---------|
| Qwen 1   | [ ]    | [ ]           | [ ]            | [ ]   | [ ]                           |         |
| Qwen 1.5 | https://github.com/EleutherAI/gpt-neox/pull/1396 | [ ] | [ ] | [ ] | [ ] |         |
| Qwen 2   | https://github.com/EleutherAI/gpt-neox/pull/1397 | [ ] | [ ] | [ ] | [ ] |         |
| Qwen 2.5 | https://github.com/EleutherAI/gpt-neox/pull/1406 | [ ] | [ ] | [ ] | [ ] |         |
| Qwen 3   | https://github.com/EleutherAI/gpt-neox/pull/1399 | [ ] | [ ] | [ ] | [ ] |         |
| Qwen 3 Next  | [ ]  | [ ] | [ ] | [ ] | [ ] |   No dense offering. Only MoE      |
| Qwen 3.5 | [ ]    | [ ]           | [ ]            | [ ]   | [ ]                           |  Uses linear attention in some layers. I don't think we have that implemented yet     |
| Qwen 3.6 | [ ]    | [ ]           | [ ]            | [ ]   | [ ]                           |   Uses linear attention in some layers. I don't think we have that implemented yet       |

## Llama

| Series | Config | Import Script | Export Script | Tests | Works with Transformer Engine | Remarks |
|---------|--------|---------------|----------------|-------|-------------------------------|---------|
| Llama 1 | [x]    | [x]           | [x]            | [ ]   | [ ]                           |        |
| Llama 2 | [x]    | [x]           | [x]            | [ ]   | [ ]                           |         |
| Llama 3 | [ ]    | [ ]           | [ ]            | [ ]   | [ ]                           |         |

## OLMo

| Series | Config | Import Script | Export Script | Tests | Works with Transformer Engine | Remarks |
|---------|--------|---------------|----------------|-------|-------------------------------|---------|
[...]

PR #1367, opened 2026-01-06
Default to only applying QK norm over the head dimension

christianazinn:
Before this change, the default application of QK norm was to apply it to dimensions `[*, N, H]`. This caused problems with GQA, in which the K tensor is of shape `[T, B, N_kv, H]`, which doesn't match. This PR fixes that by changing QK norm to normalize over `[*, H]` by default, as in e.g. Gemma3 and the original QK norm paper, with the option to go back to using `[*, N, H]` available as `qk_layernorm_over_heads`.

PR #1408, opened 2026-07-17
OLMo2 Dense Model Training Configs

aflah02:
Adding configs to support training models in the OLMo 2 series.
Addresses part of https://github.com/EleutherAI/gpt-neox/issues/1394

### DIFF
--- a/configs/olmo2/13B.yml
+++ b/configs/olmo2/13B.yml
@@ -1,0 +1,32 @@
+# Based on https://huggingface.co/allenai/OLMo-2-1124-13B/blob/main/config.json
+
+{
+  "pipe_parallel_size": 1,
+  "model_parallel_size": 1,
+  "make_vocab_size_divisible_by": 1,
+
+  # model settings
+  "num_layers": 40,
+  "hidden_size": 5120,
+  "intermediate_size": 41472, # 13824*3 as NeoX requires 3x with Swiglu
+  "num_attention_heads": 40,
+  "num_kv_heads": 40,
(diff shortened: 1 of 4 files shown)


Quentin-Anthony:
This doesn't match the HF impl of olmo2, which does RMSNorm of Q and K separately and after projection. 

See https://github.com/huggingface/transformers/blob/v5.15.1/src/transformers/models/olmo2/modeling_olmo2.py

also the norm placement order is changed from neox to olmo2. They do something like:

```
residual = x
x = Attention(x)
x = RMSNorm(x)
x = residual + x

residual = x
x = MLP(x)
x = RMSNorm(x)
x = residual + x
```

but we do:
```
x = x + attn(ln1(x))
x = x + mlp(ln2(x))
```

aflah02:
Thanks for catching this!

image

Src: https://magazine.sebastianraschka.com/p/the-big-llm-architecture-comparison

How about adding an arg called `norm_placement`?

It can be one of `pre_norm`, `post_norm_outside_residual` and `post_norm_inside_residual` accounting for the three styles in the diagram above

aflah02:
Re the QK Norm thingy, there was a discussion here - https://github.com/EleutherAI/gpt-neox/pull/1367
I can make changes to that PR and prepare it to land together with this to account for varying types of QK Norms

aflah02:
Do we also need a tensor parallel aware QK Norm to allow for things like RMS Norm to be reductions across all ranks?

PR #1414, opened 2026-08-11
Abstraction for Intel GPUs

jahatef:
Creates an abstraction for Intel GPUs. all `torch.cuda` or `torch.xpu` calls will instead call through `device_backend.py`. Implementations for different device functions can be created instead of relying on all of the functionality within `torch.cuda` to be implemented in `torch.xpu`. Profiling options that aren't available for Intel are simply ignored.

### DIFF
--- a/configs/prof.yml
+++ b/configs/prof.yml
@@ -4,8 +4,8 @@
   "profile": true,
 
   # pytorch profiler options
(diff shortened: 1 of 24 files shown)


jahatef:
I've tested training on Intel and Nvidia systems.

Issue #1405, opened 2026-07-15
Add Inter Document Attention Masking


aflah02:
Nice to have but missing feature from the library

PR #1416, opened 2026-09-02
Add Inter Doc Attn Masking

aflah02:
Adds support for #1405 

The inter-document attention masking implementation is now ready for review.

This PR:

- Adds the inter_document_attention_masking configuration option, disabled by default.
- Generates cu_seqlens and max_seqlen from document boundaries in packed GPT-2 samples.
- Uses variable-length FlashAttention for the native attention path and THD attention for BF16 Transformer Engine (te_mha).
- Prevents attention from crossing document boundaries without constructing a dense block-diagonal mask.
- Resets learned, rotary, and ALiBi positional information at each document boundary.
- Carries packed-sequence metadata through pipeline stages and activation checkpointing.
- Preserves the existing behavior when the feature is disabled.

The implementation currently supports dataset_impl: gpt2 with normal training and either native FlashAttention or BF16 Transformer Engine attention. FP8 Transformer Engine attention, sequence parallelism, curriculum sequence-length training, and soft-prompt tuning are not currently supported.

I’ve added unit tests covering dataset metadata generation, batching and loss masking, pipeline transport, activation checkpointing, positional encodings, tensor-parallel head partitions, and forward/backward equivalence against per-document reference attention.

The CI was also broken and tests wouldn't run so I've made some changes to the same


aflah02:
I've done some tiny runs and it works + tests pass.
I need to do a big run once I find some free GPU and time

aflah02:
I tested via these 4 runs ([Wandb Report](https://wandb.ai/aflah/gpt-neox-inter-document-attention-long/reports/Inter-Doc-Attn-Masking-NeoX--VmlldzoxNzg2MjEwMg?accessToken=4nm7mqbs74cb1csy1kefkw4554kuwsw2uw8yt8g1pc8c4uz25h8hhydd2enp3pto)) -

Note: Content below is heavily AI generated but I verified everything. It just was work ripe for AI use

| Run | Model implementation | Attention layout/backend | Document boundaries |
|---|---|---|---|
| Native, masking off | GPT-NeoX attention | FlashAttention fixed-shape `BSHD` during training | Ignored; tokens can attend across documents |
| Native, masking on | GPT-NeoX attention | FlashAttention variable-length `[T,H,D]` | Enforced using real `cu_seqlens` |
| TE, masking off | TE `MultiheadAttention` | FlashAttention fixed-shape `SBHD` because `NVTE_FUSED_ATTN=0` | Ignored |
| TE, masking on | TE `MultiheadAttention` | FlashAttention variable-length `THD` | Enforced using real `cu_seqlens` |

Important details:

- “Native” means the GPT-NeoX attention wrapper, not dense PyTorch attention. The base config explicitly requests `"flash"` attention.
- Native masking-on calls `flash_attn_varlen_func` directly.
- TE masking-off originally selected cuDNN FusedAttention and crashed because `head_dim=256`. The current fallback disables only cuDNN fused attention, so TE selects FlashAttention—not its unfused implementation.
- TE masking-on switches TE to `THD`. cuDNN 9.0 does not provide TE’s fused THD path, so TE selects FlashAttention 2.5.6, which supports `head_dim=256`.

All four runs train the same 1.012B-parameter GPT-NeoX model. Attention implementation and document-boundary handling are the only architectural/runtime differences.

### Model architecture

- 16 transformer layers
- Hidden size: 2,048
- 8 attention heads
- Head dimension: 256
- MLP width: 8,192 (`4 × hidden_size`)
- GELU activation
- LayerNorm with epsilon `1e-5`
- GPT-J-style parallel residual blocks
- Context length: 2,048 tokens
- Rotary embeddings on 25% of each head: 64 of 256 dimensions
- GPT-2 vocabulary padded from 50,257 to 50,304
- Input and output embeddings are not tied
- Attention and hidden dropout: `0`
- Total parameters: `1,011,781,632`

### Training configuration

- Precision: BF16 only
- FP32 gradient accumulation and all-reduce
- Microbatch size: 4
- Gradient accumulation: 8
- Effective global batch: 32 sequences
- Tokens per optimizer step: `32 × 2,048 = 65,536`
- Training iterations: 500
- Approximately 32.8 million token presentations
- Activation checkpointing: every transformer layer
- Tensor, pipeline and sequence parallelism: disabled
- DeepSpeed ZeRO stage 0
- One H100 GPU per run

### Optimizer and schedule

- Adam
- Learning rate: `2.5e-4`
- Minimum learning rate: `2.5e-5`
- Betas: `(0.9, 0.95)`
- Epsilon: `1e-8`
- Weight decay: `0.1`
- Gradient clipping: `1.0`
- Cosine decay across 500 iterations
- 1% warmup: approximately 5 iterations
- `small_init` model initialization
- `wang_init` output-layer initialization

### Data and evaluation

- WikiText-2 raw
- GPT-2 BPE tokenizer
- Memory-mapped indexed dataset
- Packed 2,048-token samples
- Train/validation/test split: `90/5/5`
- Validation every 50 iterations for 10 batches
- Checkpoints every 100 iterations and at iteration 500
- Seed: `1234`

## Results

All four 1.012B runs completed 500 iterations without skipped or NaN iterations.

Speed is averaged over logged training steps 50–500, excluding evaluation and checkpoint stalls.

| Run | Train loss | Final validation loss / PPL | Test loss / PPL | Iteration | Samples/s | Tokens/s |
|---|---:|---:|---:|---:|---:|---:|
| Native off | 3.1310 | 5.5605 / 259.95 | 5.7004 / 298.98 | 2.230 s | 14.399 | 29.5k |
| Native on | **2.8985** | 5.7597 / 317.25 | 5.9350 / 378.06 | 2.076 s | 15.449 | 31.6k |
| TE off | 3.2540 | **5.5106 / 247.30** | **5.6480 / 283.73** | 2.097 s | 15.326 | 31.4k |
| TE on | 2.9332 | 5.7528 / 315.07 | 5.9136 / 370.02 | **2.028 s** | **15.847** | **32.5k** |

### Performance

TE masking-on was fastest:

- 10.1% higher throughput than native masking-off.
- TE was 6.4% faster than native with masking off.
- TE was 2.6% faster than native with masking on.
- Enabling masking improved native throughput by 7.3%.
- Enabling masking improved TE throughput by 3.4%.

Packed variable-length attention computes attention independently over documents, reducing work from approximately `S²` to `Σ document_length²`. The linear and MLP layers still process every token, so the overall gain remains modest.

Peak allocated GPU memory was effectively identical—approximately 23.1 GB for every run. FlashAttention and activation checkpointing dominate the memory behaviour, so packing did not produce a visible model-level memory reduction.

### Loss behaviour

The masked native and TE runs tracked each other closely:

- Final validation difference: `0.0069`
- Test difference: `0.0215`
- Final training difference: `0.0347`

That close agreement is useful evidence that the TE packed path behaves consistently with the native implementation.

Masking-on produced lower training loss but worse validation and test loss in both implementations. It also began overfitting sooner:

| Run | Best validation loss | Iteration |
|---|---:|---:|
| Native off | 5.5005 | 450 |
| Native on | 5.6202 | 350 |
| TE off | **5.4845** | 450 |
| TE on | 5.6308 | 400 |

Given that this is a 1B model repeatedly training on tiny WikiText-2 with one seed, these quality differences should not be treated as a general conclusion about inter-document masking. The run is much stronger as a correctness and execution-path test than as a modeling-quality benchmark.

aflah02:
I've also kicked off some runs with TP=2, DP=2, PP=2 and sequential attention (Llama 1B based) to test the same 4 configs as above but with different model parallelism schemes. Will add results as soon as they're done. The code for these rests in https://github.com/aflah02/Inter_Doc_Attn_Mask_NeoX_Testing

aflah02:
> I've also kicked off some runs with TP=2, DP=2, PP=2 and sequential attention (Llama 1B based) to test the same 4 configs as above but with different model parallelism schemes. Will add results as soon as they're done. The code for these rests in https://github.com/aflah02/Inter_Doc_Attn_Mask_NeoX_Testing

None of the configs crash and all of them run well. I did not run long runs with these as it would take too long

PR #1424, opened 2026-09-09
Improving QK Layernorm

aflah02:
Builds on top of #1367 
Necessary for #1408


aflah02:
Ran several config tests
See report: [Wandb Report](https://wandb.ai/aflah/qk_norm_testing_remnants/reports/QK-Norm-Testing--VmlldzoxNzkwMjEwMg?accessToken=iuf4ju7kr8fmfn3k54i46lfy01n7lbwsxl44klm117hlz4t0gnzgimhq4kylowgj) (All combinations ran)
Configs Tested: https://github.com/aflah02/QKNormTesting-NeoX

aflah02:
@Quentin-Anthony What do you think about supporting a QK-Norm variant that performs a global RMSNorm across all heads when `across_heads` is enabled?

The current `across_heads` behavior is still rank-local under tensor parallelism, since each TP rank only normalizes over the heads assigned to that rank.

For OLMo 2 compatibility, it would be useful to have an AI2-style variant where the RMS statistic is computed over the full Q/K projection across all heads, even when those heads are sharded across TP ranks. AI2’s original OLMo 2 implementation gets this behavior naturally because it uses FSDP rather than Megatron-style tensor parallelism, so each rank sees the full Q/K activation when applying RMSNorm.

Under TP, reproducing that behavior would require a cross-rank reduction of the RMS statistics when `across_heads` is selected.

One straightforward implementation would be to do **two reductions per layer**:

* compute the local sum of squares for Q, all-reduce it across the TP group, and use the global statistic for `q_norm`;
* do the same independently for K and `k_norm`.

Since only the RMS statistics need to be reduced, the communication volume should be small, although this does introduce two additional synchronization points per layer.

Another option would be to reduce Q and K statistics together with **one collective per layer**. For example, compute the local Q and K sum-of-squares, pack them into a single tensor, perform one all-reduce over the TP group, and then split the resulting global statistics to normalize Q and K independently.

That should preserve separate Q/K RMSNorm semantics while reducing the additional collective count from two to one per layer.

aflah02:
Note: The above description is for RMSNorm. Other norms may reduce other sets of fields like Layernorm will requires global sums to calculate the mean - Q sum, Q squared-sum, K sum, and K squared-sum. These can still be packed into one collective.

aflah02:
I also think the name is a bit misleading and we should rename the arg to `qk_norm` and not qk_layernorm as that implies that it is layernorm to some degree when it can be other norm types too

aflah02:
I also have an initial implementation now (still needs to be tested well) for how I envision it to look like

Issue #1420, opened 2026-09-05
Several Unused Args


aflah02:
Several args are unused and are also set in multiple ways. A preliminary report is present here: https://gist.github.com/aflah02/6d895fc80b0aa7d758040deeebc61905
I will verify all the claims and overall multiple PRs close this issue

PR #1428, opened 2026-09-15
Fix FP16/BF16 configuration handling and document mixed-precision options

aflah02:
Fixes part of https://github.com/EleutherAI/gpt-neox/issues/1420 - Fix mixed-precision argument handling and remove obsolete top-level loss-scaling arguments.

## Changes

- Validate supported FP16 and BF16 dictionary keys.
- Reject ignored legacy keys such as `fp16.fp16` and `fp16.type`.
- Reject FP16 loss-scaling options inside `bf16`.
- Detect conflicts between `precision` and enabled precision dictionaries.
- Preserve dictionary-only selection through `fp16.enabled` and `bf16.enabled`.
- Normalize `precision` casing.
- Remove unused top-level `hysteresis`, `loss_scale`, `loss_scale_window`, `min_scale`, and `dynamic_loss_scale`.
- Align the Poetry DeeperSpeed revision with `requirements.txt`.'

## Notes

- Complete FP16 defaults are added only to the generated DeepSpeed configuration, leaving the user-facing `NeoXArgs.fp16` dictionary unchanged.
- Defaults are materialized because DeepSpeed otherwise has inconsistent dynamic-scaling fallbacks: fused, unfused, and ZeRO optimizers use different initial scales, minimum scales, and hysteresis values.
- `hysteresis` remains supported inside `fp16` because ZeRO uses it, although non-ZeRO FP16 wrappers do not.
- BF16 loss-scaling options are rejected because BF16 always uses a loss scale of 1.


aflah02:
Needs both more testing as well as discussion over the scope

PR #1430, opened 2026-09-15
Fix unused μP attention and output temperature arguments

aflah02:
Fixes part of #1420

### Summary

  `mup_attn_temp` and `mup_output_temp` were exposed as configuration options but did not reliably affect training.

  This PR:

  - Restores `mup_attn_temp` scaling for global attention, including when query-key layer scaling is disabled.
  - Applies `mup_output_temp` to both tied and untied language-model outputs.
  - Applies both settings only when `use_mup` is enabled.
  - Rejects non-default `mup_attn_temp` values for unsupported attention backends.
  - Validates that enabled μP temperatures are greater than zero.
  - Clarifies the argument documentation.

  The default value of `1.0` preserves existing behavior


Quentin-Anthony:
I think this is a good opportunity to tie up some loose ends about our mup that have bugged me.

One issue with our old mup impl is that we would scale the mup attention scale by the layer number due to query-key layer scaling. We should also not allow the combination of mup with flash, sparse/local attn, or TE since we don't pass the scaling factors properly yet on those. 

I also added some vibed regression tests.

If you're fine with my cleanup here, I'm good to merge.

Issue #1432, opened 2026-09-18
Fix ineffective gradient clipping caused by duplicate arguments and mismatched defaults


aflah02:
GPT-NeoX currently exposes both `clip_grad` and `gradient_clipping`, although only `gradient_clipping` is consumed by DeepSpeed.

There are two related problems:

1. `clip_grad` is unconditionally overwritten by `gradient_clipping`, so configuring it has no effect.
2. NeoX defines `gradient_clipping` with a default of `1.0` and omits default-valued options from the generated DeepSpeed configuration. DeepSpeed’s actual default is `0.0`, meaning disabled.

Consequently, the commonly used configuration:

```yaml
gradient_clipping: 1.0
```

is omitted from the generated DeepSpeed config, causing gradient clipping to remain disabled.

Expected behavior:

- `gradient_clipping: 1.0` should be passed to DeepSpeed and enable clipping.
- Omitting `gradient_clipping` should use DeepSpeed’s default of `0.0`.
- There should be only one gradient-clipping configuration option.

Proposed fix:

- Change NeoX’s `gradient_clipping` default from `1.0` to `0.0`.
- Remove the unused `clip_grad` argument and synchronization logic.
- Update the generated argument documentation.
- Add tests verifying that an explicitly configured value of `1.0` reaches DeepSpeed.

aflah02:
More explanation for the DS arg part (AI Generated but validated) - 

The problem came from how NeoX constructs the configuration dictionary that it hands to DeepSpeed.

## Configuration flow

```text
YAML file
   ↓
NeoXArgs dataclass
   ↓
Generated DeepSpeed dictionary
   ↓
DeepSpeedConfig
   ↓
Training engine
```

Suppose the YAML contains:

```yaml
gradient_clipping: 1.0
```

### 1. NeoX loads the value

Before this fix, NeoX declared:

```python
gradient_clipping: float = 1.0
```

After loading the YAML, both the current value and NeoX’s declared default were therefore identical:

```text
Current value: 1.0
NeoX default:  1.0
```

At this point, NeoX does not retain whether the value was explicitly provided by the user. It only knows that the resulting value is `1.0`.

### 2. NeoX generates a minimal DeepSpeed config

The `deepspeed_config` property requests only non-default values:

```python
config = self.get_parent_class_value_dict_extra_ds(
    NeoXArgsDeepspeedConfig,
    only_non_defaults=True,
)
```

That helper effectively performs:

```python
[...]

PR #1433, opened 2026-09-18
Fix gradient clipping configuration and remove `clip_grad`

aflah02:
Addresses #1432 and part of #1420

## Summary

Fix gradient clipping configuration by:

- Changing NeoX’s `gradient_clipping` default from `1.0` to `0.0`, matching DeepSpeed.
- Removing the unused `clip_grad` argument and synchronization logic.
- Updating the generated argument documentation.
- Adding regression tests for the corrected behavior.

Previously, `gradient_clipping: 1.0` matched NeoX’s default and was omitted from the generated DeepSpeed configuration. DeepSpeed then used its own default of `0.0`, leaving gradient clipping disabled.

With this change, explicitly configuring:

```yaml
gradient_clipping: 1.0
```

correctly passes the value to DeepSpeed.


Quentin-Anthony:
@aflah02 can you take a brief look at my latest commit and make sure you're happy with my changes? I don't want old configs with `clip_grad` to fail, so I instead am discarding it with a warning to the user. I left this as a general function for discarding potential future deprecated keys. I also updated the test to be future-proof on deepspeed clipping defaults + tested the legacy config discarding.

aflah02:
Hi @Quentin-Anthony 
I am open to the deprecation route but one counter argument I have from anecdotal experience is that a lot of times there's hundreds of lines of logs and you just ignore it and then later in the run you realize something hidden there. So if you really wanted gradient clipping maybe having it error out is better as then you'll trace to the root cause of the error and fix your run?
That said people should read warning logs so I'm also fine with your proposed changes
Once this PR is merged, I can rebase the other ones on it to reuse the function

Issue #1434, opened 2026-09-20
Support explicit attention `head_dim` independent of `hidden_size / num_attention_heads`


aflah02:
Disclaimer: The below text was largely written by ChatGPT after a long discussion with me and was independently verified by me.

GPT-NeoX currently assumes that the dimension of each attention head is always:

```python
head_dim = hidden_size // num_attention_heads
```

This forces the concatenated query-head width to equal the residual-stream width:

```text
num_attention_heads * head_dim == hidden_size
```

That relationship is common, but it is not required by multi-head attention and is not true for several current architectures. In particular, the official Qwen3-0.6B, Qwen3-4B, and Qwen3-32B configurations specify `head_dim: 128` even though `hidden_size / num_attention_heads` is respectively 64, 80, and 80.

GPT-NeoX should support an explicit per-head dimension, falling back to the current derivation when it is not configured.

This is required before the configurations proposed in [#1399](https://github.com/EleutherAI/gpt-neox/pull/1399) can faithfully represent all Qwen3 dense models.

## Background

These dimensions describe different parts of the transformer:

- `hidden_size` is the width of the residual stream entering and leaving each transformer block.
- `num_attention_heads` is the number of query heads.
- `head_dim` is the width of each query, key, and value head.
- `num_kv_heads` is the number of key/value heads used by grouped-query attention.

An attention implementation can use projections with these shapes:

```python
q_proj: hidden_size -> num_attention_heads * head_dim
k_proj: hidden_size -> num_kv_heads * head_dim
v_proj: hidden_size -> num_kv_heads * head_dim
o_proj: num_attention_heads * head_dim -> hidden_size
```

The attention-internal width does not have to equal the residual width. The input projections can expand the representation before attention, and the output projection contracts it back before the residual addition.

Hugging Face’s Qwen3 implementation follows this design:

```python
self.head_dim = getattr(
    config,
    "head_dim",
    config.hidden_size // config.num_attention_heads,
)
```

It sizes the query projection from `num_attention_heads * head_dim`, the key/value projections from `num_key_value_heads * head_dim`, and the output projection from `num_attention_heads * head_dim` back to `hidden_size`.

See the [Qwen3 attention implementation in Transformers v4.51](https://github.com/huggingface/transformers/blob/v4.51.0/src/transformers/models/qwen3/modeling_qwen3.py#L170-L222).

The quotient is a backward-compatible default, not a constraint that an explicit `head_dim` must equal.

[...]

PR #1436, opened 2026-09-20
Add head dim support

aflah02:
Addresses #1434
WIP


megatron/model/transformer.py
```diff
@@ -337,24 +341,24 @@ def __init__(
             )  # how large the total hidden dim for each of K and V is
         else:
             self.num_kv_heads_per_partition = self.num_attention_heads_per_partition
-            self.kv_hidden_size = neox_args.hidden_size
+            self.kv_hidden_size = self.query_hidden_size
 
         if not self.gqa:
```
aflah02, line 346: ColumnParallelLinear maps residual states into one fused QKV tensor:

  input:  hidden_size
  output: [Q | K | V]

  The input remains hidden_size because attention receives residual-stream vectors.

  For MHA, Q, K, and V each have width query_hidden_size:

  output_size = 3 * query_hidden_size

  For GQA/MQA, Q has query_hidden_size, while K and V each have the smaller kv_hidden_size:

  output_size = query_hidden_size + 2 * kv_hidden_size

  Previously, NeoX assumed query_hidden_size == hidden_size, which is no longer necessarily true.

  For Qwen3-0.6B:

  hidden_size       = 1024
  query_hidden_size = 16 * 128 = 2048
  kv_hidden_size    = 8 * 128  = 1024

  fused output = 2048 + 1024 + 1024 = 4096

  The old expression produced only 1024 + 1024 + 1024 = 3072, under-sizing the query projection.

megatron/model/transformer.py
```diff
@@ -887,9 +891,9 @@ def forward(self, hidden_states, attention_mask, layer_past=None):
         # [b, np, sq, hn] --> [sq, b, np, hn]
         context_layer = context_layer.permute(2, 0, 1, 3).contiguous()
 
-        # [sq, b, np, hn] --> [sq, b, hp]
+        # [sq, b, np, hn] --> [sq, b, qp]
```
aflah02, line 894: After attention, the context tensor contains one result per local query head:

  [sq, batch, num_query_heads / TP, head_dim]

  Flattening its final two dimensions therefore produces:

  (num_query_heads / TP) * head_dim
  = query_hidden_size / TP
  = qp

  So the correct reshape is:

  new_context_layer_shape = context_layer.size()[:-2] + (
      self.query_hidden_size_per_partition,
  )

  Previously, hidden_size_per_partition worked only because query_hidden_size == hidden_size.

  For Qwen3-0.6B with TP=2:

  local context width = (16 / 2) * 128 = 1024
  hidden_size / TP    = 1024 / 2 = 512  # incorrect

  Using 512 would attempt to reshape a 1024-wide context incorrectly. After flattening to qp, the
  row-parallel output projection contracts the global query width back to the residual hidden_size.

